PR #29, opened 2018-12-23
Change main project author to “the rss2email maintainers” and drop maintainer list in setup.py

Profpatsch:
Seems fitting if we move to a community-maintained project.

### DIFF
--- a/.update-copyright.conf
+++ b/.update-copyright.conf
@@ -10,6 +10,8 @@ ignored: AUTHORS | COPYING | README | README.rst | .update-copyright.conf |
 
 [aliases]
 Aaron Swartz: Aaron Swartz <me@aaronsw.com>
+Jeff Backus <jeff@jsbackus.com>: Jeff Backus <jeff.backus@gmail.com>
+Yannik Sembritzki <yannik@sembritzki.me>: Yannik <yannik@sembritzki.me>
 
 [author-hacks]
 rss2email/config.py: Aaron Swartz |
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,3 +17,4 @@ Martin 'Joey' Schulze
 Matej Cepl
 Thiago Coutinho <root@thiagoc.net>
 W. Trevor King <wking@tremily.us>
+Profpatsch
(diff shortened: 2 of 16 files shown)


Yannik:
I'd personally prefer to drop both the ever-growing AUTHORS file and __contributors__ altogether. Anyone can check the contributers in the git history.

Profpatsch:
> Anyone can check the contributers in the git history.

That’s not exactly true, the git history was cut in 2012, so e.g. Aaron Swartz (RIP) would not even appear.

But we can drop one of the two and merge them.

Yannik:
> > Anyone can check the contributers in the git history.
> 
> That’s not exactly true, the git history was cut in 2012, so e.g. Aaron Swartz (RIP) would not even appear.
> 
> But we can drop one of the two and merge them.

that sounds good to me.

Profpatsch:
Ooooh, I have made another discovery: `AUTHORS` is generated by https://pypi.org/project/update-copyright/, and the original “source of truth” here comes from the hidden file `.update-copyright.conf`.

Yannik:
Are you sure? Trevor W. King and Thiago Coutinho are in `AUTHORS` but not in `.update-copyright.conf`.

Profpatsch:
Yeah, the package generates the authors from the git history of single files, and `authors-hacks` is only used for authors that are not recorded in the git history (which started later than that).

Yannik:
ah, got it.

Profpatsch:
Okay, I pushed the result of running the `update-copyright` scripts on the current repo.

Yannik:
that looks good to me. however, do we really want 20-line contributors lists in each file (for example email.py)? I'd personally prefer to simply refer users to the AUTHORS file in the individual files.

Profpatsch:
> that looks good to me. however, do we really want 20-line contributors lists in each file (for example email.py)? I'd personally prefer to simply refer users to the AUTHORS file in the individual files.

I haven’t thought about this yet, just updating the system put into place by wking. I’ll merge for now, let’s discuss this seperately.

Issue #14, opened 2018-02-15
tests fail


Profpatsch:
It looks like the golden tests are no longer formatted the same.

Pythonpath:
```
rss2email-3.9
python3.6-setuptools-38.4.0
python3.6-bootstrapped-pip-9.0.1
python3.6-feedparser-5.2.1
python3-3.6.4
python3.6-beautifulsoup4-4.6.0
python3.6-html2text-2016.9.19
python3.6-setuptools-38.4.0
python3.6-bootstrapped-pip-9.0.1
python3.6-feedparser-5.2.1
python3-3.6.4
python3.6-beautifulsoup4-4.6.0
python3.6-html2text-2016.9.19
```

```
Traceback (most recent call last):
  File "./test/test.py", line 141, in <module>
    test(dirname=this_path, force=args.force)
  File "./test/test.py", line 75, in test
    test(dirname=dirname, config_path=config_path, force=force)
  File "./test/test.py", line 100, in test
    '\n'.join(diff_lines)))
ValueError: error processing allthingsrss/1.config
--- expected
+++ generated
@@ -7,13 +7,17 @@
 Subject: Version 2.67 Released
 Date: Tue, 21 Sep 2010 16:55:07 -0000
 Message-ID: <...@dev.null.invalid>
-User-Agent: rss2email/...
+User-Agent: rss2email/3.9 (https://github.com/wking/rss2email)
 X-RSS-Feed: allthingsrss/feed.atom
 X-RSS-ID: http://www.allthingsrss.com/rss2email/?p=62
 X-RSS-URL: http://feedproxy.google.com/~r/allthingsrss/hJBr/~3/aGbf5Gefkmc/
 X-RSS-TAGS: Uncategorized
 
-Version 2.67 of rss2email is now available for both [Linux](http://www.allthingsrss.com/rss2email/rss2email-2.67.tar.gz) and [Windows](http://www.allthingsrss.com/rss2email/rss2email-2.67.zip), which includes the latest development version of feedparser. Changes from the previous version:
+Version 2.67 of rss2email is now available for both
+[Linux](http://www.allthingsrss.com/rss2email/rss2email-2.67.tar.gz) and
+[Windows](http://www.allthingsrss.com/rss2email/rss2email-2.67.zip), which
+includes the latest development version of feedparser. Changes from the
+previous version:
 
   * Fixed entries that include an id which is blank (i.e., an empty string) were being resent
   * Fixed some entries not being sent by email because they had bad From headers
[...]

PR #30, opened 2018-12-24
Fix and trim golden tests

Profpatsch:
The tests/ folder has a few files that are basically whole feeds
checked in.
Python checked their mail module, so all of these expected files went
out of date in the mean time.

Checking in complete mailing lists with multiple thousand bytes of
redundant information is a pretty bad idea anyway, so we trim all
feeds to at most two entries. It might make sense to remove these
files completely in the future.

Fixes https://github.com/rss2email/rss2email/issues/14

Signed-off-by: Profpatsch <mail@profpatsch.de>


Profpatsch:
> ekleog | my phone being currently physically broken I can't login to github, but consider this message a +1 from me for your PR :)

Issue #25, opened 2018-12-23
Only support python >3.5


Profpatsch:
According to https://repology.org/metapackage/python3/versions current Debian Stable is on 3.5. I’d argue Debian Stable is a good reference for the oldest versions one should be supporting.

This means we can simplify parts of the code.

PR #31, opened 2018-12-24
Drop support for python 3.2, 3.3 (<3.4)

Profpatsch:
Debian Stable ships with Python 3.5 by default, and the online python
docs don’t list anything older than 3.5 as well. We can assume that
3.5 is the oldest version support by most of the ecosystem.

This removes all special casing for old APIs, as far as I can see.
It also makes automatic testing with nixpkgs possible (which ships
with support for 3.5, 3.6 and 3.7).

Fixes https://github.com/rss2email/rss2email/issues/25

Signed-off-by: Profpatsch <mail@profpatsch.de>


Profpatsch:
> so I don't see any reasons to drop 3.4 at all.

Well, the online manual doesn’t have a section for 3.4 anymore, plus cross-testing is going to be harder for 3.4. But we can drop the test in `__init__` to 3.4, no problem.

Profpatsch:
Okay, adjusted to still support 3.4

setup.py
```diff
@@ -48,8 +48,9 @@
         'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.2',
-        'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.5',
```
Yannik: can you add 3.4 here for completeness sake

Profpatsch: Left it out on purpose, because I expect us to drop it rather soon (mid-2019, when oldstable moves to current stable).

Yannik: well, everything else says 3.4. keeping this consistent and dropping it in all places is the right way to go IMO.

PR #34, opened 2018-12-28
Add nix files & a description of what they do

Profpatsch:
These are intended to ease setting up a valid development environment.
They require installing the nix package manager, so they are optional.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,6 +1,45 @@
-This file is a big TODO
+# Hacking on `rss2email`
 
-# Cutting a new release
+This document is still TODO. Please contribute to it and file issues
+if you have a question (or something is not described sufficiently).
+
+
+## Cutting a new release
 
 - TODO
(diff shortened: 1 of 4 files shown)


HACKING.md
```diff
@@ -0,0 +1,32 @@
...
+
+So far, Python 3.5, 3.6 and 3.7 are supported.
+
+You can build them like this:
+
+```
+nix-build -A rss2email-3-5 nix/release.nix
+nix-build -A rss2email-3-6 nix/release.nix
+nix-build -A rss2email-3-7 nix/release.nix
+```
```
Ekleog: Nit: add something like “You can build them all in one command with `nix-build nix/release.nix`”.

Also, the attributes would maybe be better named `rss2email-python3-5` etc.? Right now it reads as though building `rss2email`'s version `3.5` etc.

PR #36, opened 2018-12-30
[W.I.P.] Add simple CI

hartwork:


Profpatsch:
nix-based CI was added in #76

PR #37, opened 2019-01-04
Add a way to override user configured FROM field

soutade:
This commit helps to override user configured FROM field when sending an email.
It's used when we add a new feed ./r2e add <name> <url> <to> <from>
force-from configuration must be set to True

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -207,20 +207,25 @@ class Feed (object):
         'digest_post_process',
         ]
 
-    def __init__(self, name=None, url=None, to=None, config=None):
+    def __init__(self, name=None, url=None, to=None, from_email=None, config=None):
         self._set_name(name=name)
         self.reset()
         self.__setstate__(dict(
                 (attr, getattr(self, attr))
                 for attr in self._dynamic_attributes))
         self.load_from_config(config=config)
+        if not from_email is None: self.from_email = from_email
         if url:
             self.url = url
         if to:
             self.to = to
 
     def __str__(self):
-        return '{} ({} -> {})'.format(self.name, self.url, self.to)
+        if self.force_from and\
+           self.from_email != self.config['DEFAULT']['from']:
+            return '{} ({} -> {} (from {}))'.format(self.name, self.url, self.to, self.from_email)
+        else:
+            return '{} ({} -> {})'.format(self.name, self.url, self.to)
 
     def __repr__(self):
         return '<Feed {}>'.format(str(self))
@@ -319,6 +324,7 @@ class Feed (object):
         self.etag = None
         self.modified = None
         self.seen = {}
+        self.from_email = None
 
     def _set_name(self, name):
         if not self._name_regexp.match(name):
--- a/rss2email/main.py
+++ b/rss2email/main.py
@@ -88,6 +88,9 @@ def run(*args, **kwargs):
     add_parser.add_argument(
         'email', nargs='?',
         help='target email for the new feed')
+    add_parser.add_argument(
+        'from_email', nargs='?',
+        help='override from field in sent email')
 
     run_parser = subparsers.add_parser(
         'run', help=_command.run.__doc__.splitlines()[0])


Profpatsch:
If I understand this correctly, setting a custom FROM header for each feed is already supported in the config, and this just adds support for `add` to specify it on the imperative command line?

soutade:
You're right, you can set a custom FROM in the config file, but it will be
the same for all feeds.
This patch allows to set a custom FROM feed by feed.
In my case, I use it to sort feeds in different categories in my webmail.

Ekleog:
@soutade I think you can already set a custom `from =` field in feed-specific blocks, like:
```ini
[DEFAULT]
from = "foo@example.org"
# ...

[feed.feedname]
from = "bar@example.org"
# ...
```

soutade:
Ok. But you must edit the configuration file by hand, like in version 2.7 I
used. Now you can set it directly with command line during add phase.

rss2email/main.py
```diff
@@ -88,6 +88,9 @@ def run(*args, **kwargs):
     add_parser.add_argument(
         'email', nargs='?',
         help='target email for the new feed')
+    add_parser.add_argument(
+        'from_email', nargs='?',
```
Profpatsch: This should be `from-email` to be consistent.

rss2email/feed.py
```diff
@@ -207,20 +207,25 @@ class Feed (object):
...
 
-    def __init__(self, name=None, url=None, to=None, config=None):
+    def __init__(self, name=None, url=None, to=None, from_email=None, config=None):
         self._set_name(name=name)
         self.reset()
         self.__setstate__(dict(
                 (attr, getattr(self, attr))
                 for attr in self._dynamic_attributes))
         self.load_from_config(config=config)
+        if not from_email is None: self.from_email = from_email
```
Profpatsch: Conditionally setting or not setting the field is kinda bad style (and might lead to bugs). Instead, I’d read the `config['DEFAULT']['from']` field here if none is set.

rss2email/feed.py
```diff
@@ -207,20 +207,25 @@ class Feed (object):
...
         if url:
             self.url = url
         if to:
             self.to = to
 
     def __str__(self):
-        return '{} ({} -> {})'.format(self.name, self.url, self.to)
+        if self.force_from and\
+           self.from_email != self.config['DEFAULT']['from']:
```
Profpatsch: If the from field is always set we can also get rid of this special casing.

rss2email/main.py
soutade: OK. I'll change it

rss2email/feed.py
soutade: I now more understand rss2email architecture.
We read default configs in self.load_from_config and after that, we override some values if they're set in parameters. So we have to do it for "from_email" field/parameter.
The coding didn't follow actual style. I'll change it.

rss2email/feed.py
soutade: Here, I want to display overrode custom email only if it's different from the one in default.
So I need to do this checks (it's not a None check).

rss2email/feed.py
```diff
@@ -207,20 +207,26 @@ class Feed (object):
...
         self.__setstate__(dict(
                 (attr, getattr(self, attr))
                 for attr in self._dynamic_attributes))
         self.load_from_config(config=config)
+        if from_email:
+            self.from_email = from_email
         if url:
             self.url = url
         if to:
             self.to = to
```
Profpatsch: I understand that the code already was like this, but why on earth would you conditionally set fields like this?
This is going to lead to badly debuggable crashes!

```
In [1]: class Foo():      
   ...:     def __init__(self, bar=None):
   ...:         if bar:
   ...:             self.bar = bar

In [3]: Foo(bar="foo").bar
Out[3]: 'foo'

In [4]: Foo().bar
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-c643d71cfbb3> in <module>()
----> 1 Foo().bar

AttributeError: Foo instance has no attribute 'bar'

File: rss2email/feed.py (hunk 3 above)
Comment (Profpatsch) [reply rss2email/feed.py]: Based on the code above, `self.from_email` is going to crash if `from_email` was not given, because then the field doesn’t exist.

File: rss2email/feed.py [hunk 5]
```diff
@@ -319,6 +325,7 @@ def reset(self):
         self.etag = None
         self.modified = None
         self.seen = {}
+        self.from_email = None
```
Profpatsch: And here it’s set to `None` again. So it’s expected to exist. Good gracious.

rss2email/feed.py
Profpatsch: I see, I didn’t read the code correctly then (to be honest I didn’t look at the beginning of `__init__`.

PR #38, opened 2019-01-08
Feed id may be present but is empty

soutade:
Sometimes, id is present after feed parsing but remains empty, leading to empty history on saving.
So we fetch all feeds again and again when there is a new one.


Profpatsch:
Thanks!

PR #42, opened 2019-01-17
Fix inconsistency bug introduced in b0398b6bf2309ab6

soutade:
Fix inconsistency bug introduced in b0398b6bf2309ab65125566a8ee7167c840dcbcc : args set variable "from-email" and not "from_email" leading to undefined variable during feed creation.

Now --from-email is an optional argument and does not affects positional argument parsing.

### DIFF
--- a/rss2email/main.py
+++ b/rss2email/main.py
@@ -89,7 +89,7 @@ def run(*args, **kwargs):
         'email', nargs='?',
         help='target email for the new feed')
     add_parser.add_argument(
-        'from-email', nargs='?',
+        '--from-email', dest='from_email',
         help='override from field in sent email')
 
     run_parser = subparsers.add_parser(
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -214,12 +214,12 @@ class Feed (object):
                 (attr, getattr(self, attr))
                 for attr in self._dynamic_attributes))
         self.load_from_config(config=config)
-        if from_email:
-            self.from_email = from_email
         if url:
             self.url = url
         if to:
             self.to = to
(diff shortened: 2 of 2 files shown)


rss2email/main.py
```diff
@@ -89,7 +89,7 @@ def run(*args, **kwargs):
         'email', nargs='?',
         help='target email for the new feed')
     add_parser.add_argument(
-        'from-email', nargs='?',
+        '--from-email', dest='from_email',
```
Profpatsch: `--` doesn’t look right

soutade: It's the long form of this option. Without this, subparser consider it's a positional argument fetched into "from-email" member of args, which is inaccessible in Python.
I tried both 'from-email' and dest='from_email', but it complains that dest is supplied two times.
Now we keep the old behaviour with an optional argument which is, at end, not so bad.

pintman:
Same problem for me.

Profpatsch:
Hi soutade, I didn’t have time to take a closer look at your changes, the project is still in flux.

Due to https://github.com/rss2email/rss2email/issues/43 I had to revert most of your changes so far, could you rebase all of them on this PR please?

I got the feeling the changes in these PR weren’t really tested to be honest. We don’t have a CI in place yet, so we can’t catch bugs like these without manual testing unfortunately.

soutade:
Sure. I tested the first patches I sent, but this commit (from_email -->
from-email) was added to quickly from my part.

Le dim. 3 févr. 2019 à 13:07, Profpatsch <notifications@github.com> a
écrit :

PR #44, opened 2019-02-03
Add a way to override user configured FROM field in add command

soutade:
Add an optional --from-email argument to "add" command.
In combination with force_from configuration, this allows to override FROM sender email for a specific feed.
It's usefull to automatically classify feeds.

### DIFF
--- a/rss2email/command.py
+++ b/rss2email/command.py
@@ -48,7 +48,7 @@ def email(feeds, args):
 
 def add(feeds, args):
     "Add a new feed to the database"
-    feed = feeds.new_feed(name=args.name, url=args.url, to=args.email)
+    feed = feeds.new_feed(name=args.name, url=args.url, to=args.email, from_email=args.from_email)
     _LOG.info('add new feed {}'.format(feed))
     if not feed.to:
         raise _error.NoToEmailAddress(feed=feed, feeds=feeds)
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -207,7 +207,7 @@ class Feed (object):
         'digest_post_process',
         ]
 
-    def __init__(self, name=None, url=None, to=None, config=None):
+    def __init__(self, name=None, url=None, to=None, from_email=None, config=None):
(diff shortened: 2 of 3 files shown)


Profpatsch:
Hm, now we have 

```
usage: rss2email add [-h] [--from-email FROM_EMAIL] name url [email]
```

So both `--from-email` and `email` are optional, but one is positional and the other is not.

soutade:
Yes.
I think "--from-email" won't be used a lot compared to "email".
I can turn it positional by set it "from_email", but it doesn't look like other parameters schema.
From my point of view, I don't like optional positional argument, because they became mandatory when there are multiples arguments and you only want to set the last one.

Ekleog:
@soutade I think that we either should not have this argument, or we should allow to set *every* config file parameter via arguments, for consistency, including future ones that are not defined yet. Something like `[--option NAME VAL]`. Ideally we'd remove the `[email]` positional parameter for the next breaking release.

Does that plan sound good to you?

soutade:
Sure. It's better for coherency. So, I keep my patch in my own fork.

Ekleog:
Done, thanks! Keeping this open so as to track this pending refactor of arguments.

kaashif:
Hey @soutade, great idea for a feature, this is indeed very useful. I am working on the more general version of overriding config values on the command line in #61, which lets you do `r2e -s from <my-custom-from-address>` as well as doing that with any config key. We wouldn't have got the idea for this feature without this PR, so I can add your name to a commit if you'd like.

In any case, I'd prefer discussion to move there, since I think the more general solution (inspired by @Ekleog's comment) is an improvement.

soutade:
Hello,

Sure, a more generic switch is preferable for this kind of feature. It'll
allow us to be flexible for the coming years and not have a tons of
specific options. I like it.
In the other case, it would be good to document all new available but
"hidden" features for newcomers.
You don't have to mention my name for this, github will remind me thanks to
issue #44. I may commit later some new nice features if found one :)

PR #45, opened 2019-02-07
README: add reference to freenode channel

Profpatsch:


Ekleog:
Would it be possible to just drop `README`? It sounds useless to keep both around

Profpatsch:
Done.

PR #57, opened 2019-06-19
fix: enable to debug the name of the feed that times-out

### LINKED ISSUE #51: timeout: print the name of the feed that times-out
In my dozens of feeds, one times-out:

    [ERROR] 60 second timeout exceeded

It would be great to know which one to remove or update the feed URL.

monperrus:
fix #51

### DIFF
--- a/rss2email/command.py
+++ b/rss2email/command.py
@@ -61,6 +61,8 @@ def run(feeds, args):
     try:
         for index in args.index:
             feed = feeds.index(index)
+            # to debug feeds that timeout, run "r2e -V run"
+            _LOG.warning('refreshing feed {}'.format(feed))
             if feed.active:
                 try:
                     feed.run(send=args.send)
--- a/rss2email/command.py
+++ b/rss2email/command.py
@@ -61,8 +61,8 @@ def run(feeds, args):
     try:
         for index in args.index:
             feed = feeds.index(index)
-            # to debug feeds that timeout, run "r2e -V run"
-            _LOG.warning('refreshing feed {}'.format(feed))
+            # to debug feeds that timeout, run "r2e -VV run"
+            _LOG.info('refreshing feed {}'.format(feed))
             if feed.active:
                 try:
                     feed.run(send=args.send)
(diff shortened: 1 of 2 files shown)


Ekleog:
Wouldn't it be slightly better with `_LOG.info` instead of `_LOG.warning`? (that would then be viewable using more `-V`)

monperrus:
This is what I did in the first place, but `info` logs are not shown with the `-V` switch.

Ekleog:
They should be shown with `-VV` (`-VVV` being for `_LOG.debug`) :)

Profpatsch:
I’d print to stderr by default, without having to give a verbosity flag.

Ekleog:
Agreed. OTOH I think it should be `_LOG.info` and the default verbosity should be set to `info`, that would give the same end-result but with proper code semantics :)

Also, it'd ideally require to change the CLI to allow reducing the log level (something like “quiet”).

What about landing this changed with `_LOG.info` that also bumps the default log level to `info`, and opening for later an issue to add a “quiet” argument for people who want only warnings and errors? (anyway, to me, the current default of do-not-display-warnings appear to be totally wrong)

monperrus:
Changed to `info` log level and bumped the default log level to info

Profpatsch:
Not `warning?

Ekleog:
I guess `info` vs. `warning` as a default is a matter of preference -- I personally prefer `info`, with no strong opinion.

However, I've just noticed https://github.com/rss2email/rss2email/blob/master/rss2email/main.py#L152 ; which looks like it's not actually increasing the log level, but resetting it from 0 -- meaning `r2e -V` and `r2e` are, AFAIU, exactly the same thing with the current `warning` default.

So the fix might also be, instead of replacing `warning` with `info` as the default, to replace `ERROR` with `WARNING` in the `args.verbose` computation, so that the first `-V` actually has an effect. Does that sound good to you?

monperrus:
> Does that sound good to you?

Yes. It's the first time I see arithmetic in log level: `_logging.ERROR - 10 * args.verbose` how does this work?

Ekleog:
@monperrus It comes from https://docs.python.org/3/library/logging.html#levels / https://docs.python.org/3/library/logging.html#logging.Logger.setLevel ; python predefines logging levels as integers that just get named aliases :)

Profpatsch:
So, should we merge?

monperrus:
I'd say yes, it's the first milestone towards better logging.

monperrus:
Thanks!

PR #58, opened 2019-07-01
Use lockf so locking data file works on NFS (fixes #10)

### LINKED ISSUE #10: Don't use flock on read-only files
While flock() will work on local filesystems when used with files opened read-only, it will fail on NFSv4.  The trace below comes from an older version of rss2email, but the bug appears to be present in the code, still.

flock() should be used only when a file has been opened for writing.

$ r2e delete 6
  File "/usr/share/rss2email/rss2email.py", line 936, in <module>
    delete(int(args[0]))
  File "/usr/share/rss2email/rss2email.py", line 858, in delete
    feeds, feedfileObject = load()
  File "/usr/share/rss2email/rss2email.py", line 495, in load
    fcntl.flock(feedfileObject.fileno(), locktype)
IOError: [Errno 9] Bad file descriptor

kaashif:
If your home is on NFS (as mine is), you get this error when we try to lock the data file:

```
  File "/home/kaashif/.local/bin/r2e", line 5, in <module>
    rss2email.main.run()
  File "/home/kaashif/.local/lib/python3.6/site-packages/rss2email/main.py", line 168, in run
    feeds.load(lock=lock)
  File "/home/kaashif/.local/lib/python3.6/site-packages/rss2email/feeds.py", line 247, in load
    self._load_feeds(lock=lock, require=require)
  File "/home/kaashif/.local/lib/python3.6/site-packages/rss2email/feeds.py", line 270, in _load_feeds
    _fcntl.flock(self._datafile_lock.fileno(), locktype)
OSError: [Errno 9] Bad file descriptor
```

This is because Python's `fcntl.flock` uses the `flock` syscall, which doesn't work on NFS. We can instead use Python `fcntl.lockf`, which uses the `fcntl` syscall, which does work over NFS. Despite the name, `lockf` in Python does not use the `lockf` C function. We can't use a `LOCK_EX` lock any more, since exclusively locking a file open only for reading (not writing) fails with a "bad file descriptor" error. We don't need an exclusive lock anyway since the file is open for reading only, a shared lock is fine.

### DIFF
--- a/rss2email/feeds.py
+++ b/rss2email/feeds.py
@@ -266,8 +266,8 @@ class Feeds (list):
 
         locktype = 0
         if lock and UNIX:
-            locktype = _fcntl.LOCK_EX
-            _fcntl.flock(self._datafile_lock.fileno(), locktype)
+            locktype = _fcntl.LOCK_SH
+            _fcntl.lockf(self._datafile_lock.fileno(), locktype)
 
         self.clear()
 


Profpatsch:
> If your home is on NFS (as mine is)

oh what. Why would you do that in 2019.

kaashif:
It's still 1995 in my house, judging by the number of Sun and HP workstations lying around. And there actually is a check for SunOS in the rss2email source code which disables locking entirely, which I find weird.

I just like being a freak, leave me alone :)

PR #60, opened 2019-07-04
Add sleep interval between fetches from same server (closes #53)

### LINKED ISSUE #53: feature: Pause between requests
I have run into an issue where my requests are denied by server apparently due to hitting the server too frequently.   For this setup, I have over 60 feeds configured that all point to the same domain/server.   So when r2e is run, all those fetches occur in a loop as fast as possible.    The server sees this as abuse and rate limits, in this case returning 403 forbidden.

I would suggest a new config option: same-server-fetch-interval

Or if that's too complicated to implement, a simpler thing would be just a pause interval between each feed request, regardless of domain/server.

kaashif:
This adds the `same-server-fetch-interval` config key, which lets the user specify a sleep interval between consecutive fetches from the same server. This would fix the problem described in #53. I think it is generally a good idea to avoid being an obnoxious client, and this contributes to that.

I'm not convinced I added the config key correctly, I think someone should look this over - I have only a vague understanding of the code. I have tested it and it seems to work though.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -9,6 +9,9 @@ import io as _io
 import logging as _logging
 import os as _os
 import re as _re
+import tempfile
+import multiprocessing
+import subprocess
 
 import rss2email as _rss2email
 import rss2email.config as _rss2email_config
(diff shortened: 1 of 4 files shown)


Profpatsch:
Can we add a simple integration test for this with the builtin python http library? Something that sets the interval to 1s client-side and asserts that there’s at least 0.8s (or something) between requests.

kaashif:
If that is wanted, then shouldn't there also be wider integration tests covering the SMTP and IMAP code using e.g. [localmail](https://pypi.org/project/localmail/)? Currently the tests mess around with the guts of the Feed class so the email sending code is not tested. Maybe I should write tests for the existing network client code, get them merged (in a different PR), and only after that write a test for this and merge this PR. That is, after I have already made a place to put networky integration tests.

What I really want to do is use `unittest` to organise the tests (the existing tests included) a bit so a developer can more easily separate the setup/teardown (this is more relevant for networking tests since we have to start a server, run tests, then shut down the server) and we get nice test reports. As it is, the tests printing no output on success is a bit disconcerting, I like to see a big OK and see how many tests ran.

I want to work on this and submit a PR, but I'd like to know if anyone else also wants this.

Profpatsch:
> If that is wanted, then shouldn't there also be wider integration tests covering the SMTP and IMAP code using e.g. [localmail](https://pypi.org/project/localmail/)?

Probably, but we should start adding tests for new features we add; no need to add tests for everything else first.

kaashif:
I wrote a test checking if the delay works correctly. I think there should be more integration tests like I said, but that's another PR for another day :)

test/test.py
```diff
@@ -99,6 +102,61 @@ def test(dirname=None, config_path=None, force=False):
...
+    num_requests = 3
+
+    def webserver(queue):
+        import http.server, time
+
+        class NoLogHandler(http.server.SimpleHTTPRequestHandler):
+            def log_message(self, format, *args):
+                return
+
+        httpd = http.server.HTTPServer(('', 8000), NoLogHandler)
```
Profpatsch: Can you instead put a random unused port here (passing `0` usually does the trick), so that the test doesn’t fail if a user has something listening on 8000.

test/test.py
```diff
@@ -99,6 +102,61 @@ def test(dirname=None, config_path=None, force=False):
...
+        start = 0
+        for _ in range(num_requests):
+            httpd.handle_request()
+            end = time.time()
+            if end - start < 1:
```
Profpatsch: A second is a bit long for an automated test, something <300ms should be fine.

test/test.py
kaashif: Good catch, I just addressed this.

test/test.py
kaashif: Good suggestion, I changed the test to wait 0.3s. I also needed to change the `same-server-fetch-interval` to be a float rather than an int to wait less than a second.

PR #61, opened 2019-07-12
Add -s/--set command line option to set config keys on the command line

kaashif:
This is an implementation of the more general config overriding @Ekleog mentioned in the comments on #44 .

The behaviour is that if you override a key on the command line, it is only overridden for the duration of that invocation - your overriden value is *not* added to the config file. This is reasonable behaviour and mirrors what git does, for example.

The exception is when the in-memory config has a section added to it e.g. when you are adding a new feed (actually I think this is the only time that happens). Then, the overrides *are* written to disk in the new section. This means the user can do things like overriding the "to" address for a feed when adding it, by invoking:

```
$ r2e -s to kaashif@example.com add eff https://eff.org/rss/updates.xml
```

And this "to" address will be written to disk in the config file. In all other cases we are not adding a new section to the config, so the config overrides do not persist.

In terms of the implementation, I think I've done it in the way that involves the fewest changes to the source code. Yes it does involve a global variable in `config.py` but I think this is the easiest way to make sure the overrides are consistently used everywhere even if the config is re-read from disk and we have a new Feeds object. I'm not sure whether this happens anywhere though, maybe a cleaner design would also work.

Comments on the design, naming, implementation etc are welcome.

### DIFF
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -40,12 +40,34 @@ import configparser as _configparser
 
 import html2text as _html2text
 
+class SectionOverrideProxy(_configparser.SectionProxy):
+    def __init__(self, name, config, overrides={}):
+        super().__init__(config, name)
+        # Get the real (not affected by overrides) section
+        self.real_section = _configparser.ConfigParser.__getitem__(config, name)
+        self.overrides = overrides
+
+    def __getitem__(self, key):
+        if key in self.overrides:
+            return self.overrides[key]
+        else:
+            return self.real_section[key]
+
 class Config (_configparser.ConfigParser):
     def __init__(self, dict_type=_collections.OrderedDict,
                  interpolation=_configparser.ExtendedInterpolation(),
                  **kwargs):
         super(Config, self).__init__(
             dict_type=dict_type, interpolation=interpolation, **kwargs)
+        self.overrides = {}
+        self.filenames = None
+
+    def __getitem__(self, section_name):
+        """Show a section with overrides enabled. This is usually what is
+        wanted, since the user can configure the program with
+        overrides on the command line (via -s/--set).
+        """
+        return SectionOverrideProxy(section_name, self, self.overrides)
 
     def setup_html2text(self, section='DEFAULT'):
         """Setup html2text globals to match our configuration
@@ -61,6 +83,36 @@ class Config (_configparser.ConfigParser):
             section, 'links-after-each-paragraph')
         _html2text.BODY_WIDTH = self.getint(section, 'body-width')
 
+    def set_overrides(self, overrides):
+        self.overrides = overrides
+
+    def read(self, filenames):
+        # if we read in from a file, we need to keep track of the
+        # filenames that produced this config for later comparison, so
+        # we don't overwrite disk configuration with command-line
+        # overrides except in some specific cases
+        read_filenames = super(Config, self).read(filenames)
+        self.filenames = read_filenames
+        return read_filenames
+
+    def write(self, f):
+        """Writes config to file. Overrides are writen to disk only if the
+        section being overridden does not exist on disk (e.g. when
+        creating a new feed)"""
+        if self.filenames:
+            # We only need to take care if there is an on-disk config
+            # to worry about preserving
+            disk_config = _configparser.ConfigParser()
+            disk_config.read(self.filenames)
+            new_sections = set(self.sections()).difference(set(disk_config.sections()))
+            for section in new_sections:
+                # Create the sections with the overrides
+                disk_config[section] = dict(self[section])
+            disk_config.write(f)
+        else:
+            # Else we are creating the on-disk config for the first
+            # time and can just write everything
+            super(Config, self).write(f)
 
 CONFIG = Config()
 
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -108,6 +108,10 @@ class Config (_configparser.ConfigParser):
             for section in new_sections:
                 # Create the sections with the overrides
                 disk_config[section] = dict(self[section])
+
+            # Change the default target email if we have changed it
+            # (this only happens through 'r2e email')
+            disk_config['DEFAULT']['to'] = self['DEFAULT'].real_section['to']
             disk_config.write(f)
         else:
             # Else we are creating the on-disk config for the first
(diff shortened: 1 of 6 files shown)


Profpatsch:
Will try to review later today or tomorrow.

kaashif:
I would still like someone to review this after I fix the conflicts, which I might get time to do soon.

Profpatsch:
If you rebase it now we have CI.

kaashif:
Merged everything, tests pass (but don't take their word for it). I think the changes are pretty small and easy to review though.

rss2email/config.py
```diff
@@ -40,6 +40,22 @@
 
 import html2text as _html2text
 
+# These key/value pairs override all values with the same key in all
+# sections, including feed-specific sections. Overrides can be set
+# with the -s/--set command line option.
+OVERRIDES = {}
```
Profpatsch: Does this have to be a global variable? Can’t we use the normal data path?

rss2email/config.py
```diff
@@ -48,6 +64,18 @@ def __init__(self, dict_type=_collections.OrderedDict,
...
+        """Show a section with overrides enabled. This is usually what is
+        wanted, since the user can configure the program with
+        overrides on the command line (via -s/--set).
+        """
+        return SectionOverrideProxy(section_name, self)
+
+    def get_real_section(self, section_name):
+        """This function shows the section as it appears on disk.
+        """
+        return super().__getitem__(section_name)
```
Profpatsch: Exposing this is a breach of abstraction really; if you go through all the pain of setting up a proxy class, the class should contain all logic related to merging of configs and stuff.

kaashif:
Made some changes:

1. After reading the code some more, I was able to understand how to make the overrides not global

2. I put all of the config merging/overriding logic into `config.py`. The issue was that if you just override `__getitem__` on the section proxy but then call `.write()` on the parent config, the parent config doesn't care what the proxy thinks, so always writes the non-overridden value. We need to override .write on the config itself. And for that to work, we need to know whether a section is new (so fair game to write to disk with overrides since it's not already on disk), we need to override `.read()` to keep track of what file the config came from so we can compare when the time comes to write to disk.

I think this makes more sense and is less brittle than before, since writes are now explicitly overridden and made safe. Bonus: there are no code changes in the code that actually uses the config reading/writing logic.

rss2email/config.py
kaashif: Good point, made change to correct this.

rss2email/config.py
kaashif: Yes, I think it is in fact the `Config` object that should manage the merging of configs when written since all writes go through the `Config.write()` method anyway, so we can just override that.

kaashif:
Also, overriding writes to the config means we now have to explicitly allow specific changes to the user's config file. Previously, the program read the config, messed around with it, then wrote the whole thing to disk, with no protection against messing up user settings. Now we explicitly only allow new sections and changes to the default to address (which is done through `r2e email`).

rss2email/config.py
```diff
@@ -61,6 +83,36 @@ def setup_html2text(self, section='DEFAULT'):
             section, 'links-after-each-paragraph')
         _html2text.BODY_WIDTH = self.getint(section, 'body-width')
 
+    def set_overrides(self, overrides):
+        self.overrides = overrides
```
Profpatsch: This should be set in the constructor if possible.

rss2email/main.py
```diff
@@ -62,6 +62,9 @@ def run(*args, **kwargs):
     parser.add_argument(
         '-V', '--verbose', default=0, action='count',
         help='increment verbosity')
+    parser.add_argument(
+        '-s', '--set', action='append', nargs=2, metavar=('key', 'value'),
+        help='override config key to value')
```
Profpatsch: I’m not sure this flag is a good global flag.

What happens with `r2e list`? Does it make sense for `r2e pause`? `unpause`? `oplimport`?

rss2email/config.py
```diff
@@ -61,6 +83,36 @@ def setup_html2text(self, section='DEFAULT'):
...
+        # we don't overwrite disk configuration with command-line
+        # overrides except in some specific cases
+        read_filenames = super(Config, self).read(filenames)
+        self.filenames = read_filenames
+        return read_filenames
+
+    def write(self, f):
+        """Writes config to file. Overrides are writen to disk only if the
+        section being overridden does not exist on disk (e.g. when
+        creating a new feed)"""
```
Profpatsch: I’m not sure this is a good way to handle the overrides; because it will sometimes modify the on-disk config and sometimes not, and there is no clean way for the user to see when that really happens and when not. See also the other comments.

rss2email/main.py
kaashif: The only time I have found this useful so far is when using `r2e add`, so maybe we should just make it a flag for that. I can also envision someone using it with `r2e run` maybe to override the `to` address, to change the `email-protocol` temporarily or to change the mailbox the mail gets sent to, something like that.

rss2email/config.py
kaashif: The reason I've done it this way is that you can pass in a pre-existing `Config` object to the constructor of `Feeds` but we still want to apply the overrides in that case, so we have to have a way to do it after construction. But actually the `config` argument is only used in the tests, which don't use any overrides, so we actually don't need this...

rss2email/config.py
kaashif: The whole point of this feature is to enable users to set config keys on the command line to be written to disk when adding a new feed, and that is the only time overrides are written to disk, since that is the only time the program adds a new section.

kaashif:
I don't overwrite the config file when a command line config key is set, in fact the only command that overwrites the config file is `r2e email`. The overrides only ever make it to disk when using `r2e add` and they do not overwrite anything, they are added in a new section to the existing config.

rss2email/config.py
Profpatsch: It sounds overly complicated tbh :)

rss2email/main.py
Profpatsch: Let’s just add it to `add` for now and maybe think about other subcommands later.

rss2email/config.py
Profpatsch: I guess just putting it in the `add` command would help :)

kaashif:
I'm probably just never going to get the time to finish this PR.

PR #62, opened 2019-07-21
fix: clarify handling of guid in _process_entry

### LINKED ISSUE #39: Fix up the `guid` section in feed.py
https://github.com/rss2email/rss2email/blob/master/rss2email/feed.py#L453-L466

We call a function named `_get_entry_id` and then in the next line suddenly manually `guid = entry.get('id', id_)`. WTF?

I don’t get this part of the code at all. It should be refactored.

monperrus:
Just finished a really hard debug session due to guid.

It is confusing that:

* the handling of guid is spread over two methods `_process_entry` and `_get_entry_id`.
* the test for "entry already seen" mixes two strategies configured differently `if guid in self.seen` and `if self.seen[guid]['id'] == id_:`
* the debug information through `LOG.debug` and `X-RSS-ID` does not correspond to the setting in the config file

This pull request solves those three problems:

* the handling of guid is fully encapsulated in `_get_entry_id`, with no exception or overriding.
* the key in `self.seen` is the unique decision procedure: `if guid in self.seen`
* all debug information rely on the `guid` as provided by `_get_entry_id` and configured by the config file

Fix #39

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -449,21 +449,12 @@ class Feed (object):
             raise
 
     def _process_entry(self, parsed, entry):
-        id_ = self._get_entry_id(entry)
-        # If .trust_guid isn't set, we get back hashes of the content.
-        # Instead of letting these run wild, we put them in context
-        # by associating them with the actual ID (if it exists).
-        guid = entry.get('id', id_)
-        if isinstance(guid, dict):
-            guid = guid.values()[0]
-        # In some bad RSS feeds, id is present but empty...
-        if guid == '':
-            guid = id_
+        # get_entry_id is the unique method that encapsulates where the strategy for getting the id is set
+        guid = self._get_entry_id(entry)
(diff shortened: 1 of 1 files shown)


monperrus:
That's ready to merge.

monperrus:
somebody to merge?

kaashif:
This seems like it works for me, but I have to be honest with you, I didn't really get the code before and I never really bothered trying to decipher it. This is also related to #39. I'll merge this myself in a bit after a testing it more but let me also tag @Profpatsch in case they didn't see this.

monperrus:
Indeed, it fixes #39.

FYI, works like a charm on 80 feeds for 6 weeks now.

Profpatsch:
Let’s merge then. We really need CI and a good test suite …

kaashif:
> We really need CI and a good test suite …

I couldn't agree more :)

PR #71, opened 2019-09-03
Test maildir sending

kaashif:
I restructured the tests to use `unittest` and wrote a test that runs r2e with `email-protocol = maildir` and checks the maildir for the messages. It works by actually starting a web server and actually having a real maildir on disk, there is no mocking here. I intend to do the same kind of thing with IMAP and SMTP.

I kept the old tests since they are still useful but maybe I'll rejig them to actually run r2e and not mess around with the internals so much. That will happen in a later PR.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -127,6 +127,35 @@ class TestEmails(unittest.TestCase):
             else:
                 self.run_single_test(config_path=this_path)
 
+class ExecContext:
+    """Creates temporary config, data file and lets you call r2e with them
+    easily. Cleans up temp files afterwards.
+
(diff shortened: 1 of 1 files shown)


kaashif:
I think I'll just merge this without anyone else reviewing. The only file changed is `test/test.py` so even if the new tests I added are complete crap, there's no way it can break anything - someone can just improve the tests in a later PR. Real continuous integration coming soon.

Profpatsch:
I added CI in https://github.com/rss2email/rss2email/pull/76, from now on these should be easier to review.

PR #74, opened 2019-09-13
Allow configuration of User-Agent string

anders-dc:
This PR adds the option to change the `User-Agent` string that is sent along with emails. I personally use `urlview` to filter the rss2email messages I receive, and wanted to avoid having the rss2email github link show up in every email. The new configuration parameter has the following default value:

    user-agent = rss2email/_VERSION_ (https://github.com/rss2email/rss2email)

The `_VERSION_` part is replaced with the appropriate version number, meaning that the **default behavior is unchanged**. I also added a test (`test/allthingsrss/4.{config,expected}`). Thanks for considering this PR.

### DIFF
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -40,8 +40,6 @@ import configparser as _configparser
 
 import html2text as _html2text
 
-from . import __url__
-
 class Config (_configparser.ConfigParser):
     def __init__(self, dict_type=_collections.OrderedDict,
                  interpolation=_configparser.ExtendedInterpolation(),
@@ -71,8 +69,8 @@ CONFIG['DEFAULT'] = _collections.OrderedDict((
         ### Addressing
         # The email address messages are from by default
         ('from', 'user@rss2email.invalid'),
-        # The User-Agent default string (__VERSION__ is replaced)
-        ('user-agent', 'rss2email/__VERSION__ ({})'.format(__url__)),
+        # The User-Agent default string (rss2email __VERSION__ and __URL__ is replaced)
+        ('user-agent', 'rss2email/__VERSION__ (__URL__)'),
         # Transfer-Encoding. For local mailing it is safe and
         # convenient to use 8bit.
         ('use-8bit', str(False)),
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -40,6 +40,9 @@ import configparser as _configparser
 
 import html2text as _html2text
 
+from . import __url__
+from . import __version__
+
 
 class Config (_configparser.ConfigParser):
     def __init__(self, dict_type=_collections.OrderedDict,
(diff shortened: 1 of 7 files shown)


rss2email/feed.py
```diff
@@ -219,6 +217,7 @@ def __init__(self, name=None, url=None, to=None, config=None):
             self.url = url
         if to:
             self.to = to
+        self.user_agent = self.user_agent.replace('_VERSION_', __version__)
```
Profpatsch: I’d use `__VERSION__` (two underscores) to be aligned with the python variable.

anders-dc:
Done. Thanks!

rss2email/config.py
```diff
@@ -70,6 +71,8 @@ def setup_html2text(self, section='DEFAULT'):
         ### Addressing
         # The email address messages are from by default
         ('from', 'user@rss2email.invalid'),
+        # The User-Agent default string (__VERSION__ is replaced)
+        ('user-agent', 'rss2email/__VERSION__ ({})'.format(__url__)),
```
Ekleog: Would it be possible to interpolate `__version__` in the same way as you're interpolating `__url__` here?

Ekleog:
Other things I just noticed: this needs a `CHANGELOG` entry and addition of the new parameter to the man page, so that it's discoverable :)

anders-dc:
> Hmm I was thinking the opposite actually: have the string be `'rss2email/{} ({})'.format(__version__, __url__)`

Ah, I understand now. That wouldn't work because the default value would be written to `rss2email.cfg` when running `r2e new`, and the version number (and url) at that time would be kept in the configuration indefinitely, even if r2e's version and url change.

anders-dc:
> Other things I just noticed: this needs a `CHANGELOG` entry and addition of the new parameter to the man page, so that it's discoverable :)

I see, let me get right on that.

Ekleog:
Hmm… I would have assumed it did not? It's been really long since I last ran `r2e new`, but doesn't this change just define the default values for when a key is *not* actually defined in `rss2email.cfg`? In which case it should work properly, if `r2e new` just doesn't add the `user-agent` key to the file

anders-dc:
> Hmm… I would have assumed it did not? It's been really long since I last ran `r2e new`, but doesn't this change just define the default values for when a key is _not_ actually defined in `rss2email.cfg`? In which case it should work properly, if `r2e new` just doesn't add the `user-agent` key to the file

`r2e new` actually hard-writes the parsed string to the configuration file, so I *think* this would be the better approach. But otherwise, the functionality also uses the same values for undefined parameters.

anders-dc:
Man page and changelog updated (574ba59ae7032e5d6e6a0ae77b22a06487f6c247  and 8f417cc08eca660e553ca91772a2258ccc799472)

kaashif:
You added a test that passes, updated documentation, 2 reviewers approved, and most importantly - the feature works on my machine. Good job @anders-dc, merged.

PR #76, opened 2019-09-14
Ci add travis nix config

### LINKED ISSUE #12: Cover rss2email by continuous integration?
Hi!

I don't see any traces of something like Travis CI or so here yet. Are there plans to get that layer of protection against regressions?

Best, Sebastian

Profpatsch:
This adds a basic travis CI setup.

It also fixes the code so that it works down to python 3.5 (again)

Closes #12

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -177,7 +177,8 @@ class TestFetch(unittest.TestCase):
         num_requests = 3
 
         def webserver(queue):
-            with http.server.HTTPServer(('', 0), NoLogHandler) as httpd:
+            httpd = http.server.HTTPServer(('', 0), NoLogHandler)
+            try:
                 port = httpd.server_address[1]
                 queue.put(port)
 
@@ -190,6 +191,8 @@ class TestFetch(unittest.TestCase):
                         return
                     start = end
                 queue.put("ok")
+            finally:
+                httpd.server_close()
 
         queue = multiprocessing.Queue()
         webserver_proc = multiprocessing.Process(target=webserver, args=(queue,))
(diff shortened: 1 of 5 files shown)


test/test.py
```diff
@@ -177,19 +177,20 @@ def test_delay(self):
...
+            port = httpd.server_address[1]
+            queue.put(port)
+
+            start = 0
+            for _ in range(num_requests):
+                httpd.handle_request()
+                end = time.time()
+                if end - start < wait_time:
+                    queue.put("too fast")
+                    return
```
kaashif: If the test fails because `r2e` fails to wait, then the webserver function will return here (line 190) without properly shutting down the http server, giving the following warning when the tests fail:

```
./usr/lib/python3.6/multiprocessing/process.py:93: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('0.0.0.0', 42057)>
```

Maybe wrapping the loop in a `try...finally` would be better if you really want to remove the with block

Profpatsch:
> Is there a use case for Python 3.5 that you have in mind?

I don’t see why we *wouldn’t* support 3.5, as it’s not a lot of overhead.

test/test.py
Profpatsch: Good catch.

Profpatsch: Done, see force push diff.

Profpatsch:
try/finally fixed, PTAL

Profpatsch:
I’m trying to optimize the nix store caching a bit, will merge once that works.

kaashif:
Like I said, 3.5 is 3 years old and since all distros have newer versions, I'm not sure why we _should_ maintain compatibility. Of course, since

Issue #83, opened 2019-11-27
~/.local/share/rss2email.json gets corrupted


ezbik:
Hello, I run `r2e` in Cron, and sometimes I want to run it manually. And I believe it leads to its JSON storage corruption, when 2 processes are run at the same time. Restoring the JSON file fixes it, of course.

```
# r2e -VV run  gaf 
2019-11-27 08:00:32,635 [INFO] refreshing feed gaf (https://pathos.tanatos.org/gaf.rss -> rss+gaf@localhost)
2019-11-27 08:00:32,636 [INFO] fetch gaf (https://pathos.tanatos.org/gaf.rss -> rss+gaf@localhost)
2019-11-27 08:00:33,086 [INFO] process gaf (https://pathos.tanatos.org/gaf.rss -> rss+gaf@localhost)
2019-11-27 08:00:33,088 [WARNING] incorrectly declared encoding: document declared as us-ascii, but parsed as utf-8: gaf (https://pathos.tanatos.org/gaf.rss -> rss+gaf@localhost)

...

2019-11-27 08:00:40,924 [INFO] send message for gaf (https://pathos.tanatos.org/gaf.rss -> rss+gaf@localhost)
Traceback (most recent call last):
  File "/usr/local/bin/r2e", line 4, in <module>
    __import__('pkg_resources').run_script('rss2email==3.11', 'r2e')
  File "/opt/rss2email/lib/python3.6/site-packages/pkg_resources/__init__.py", line 658, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/opt/rss2email/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1445, in run_script
    exec(script_code, namespace, namespace)
  File "/opt/rss2email/lib/python3.6/site-packages/rss2email-3.11-py3.6.egg/EGG-INFO/scripts/r2e", line 5, in <module>
  File "/opt/rss2email/lib/python3.6/site-packages/rss2email-3.11-py3.6.egg/rss2email/main.py", line 171, in run
  File "/opt/rss2email/lib/python3.6/site-packages/rss2email-3.11-py3.6.egg/rss2email/command.py", line 92, in run
  File "/opt/rss2email/lib/python3.6/site-packages/rss2email-3.11-py3.6.egg/rss2email/feeds.py", line 358, in save
[...]

PR #94, opened 2020-02-01
Lock the datafile at the start, only release at the end

kaashif:
This PR corrects how the locking works, which caused issues in e.g. https://github.com/rss2email/rss2email/issues/83. The problem is outlined in a comment I wrote there but basically, the desired behaviour is that if there are multiple instances of r2e about, they should take it in turns to run and avoid all running at the same time, possibly mangling the data file or the config file.

I implement this by acquiring an exclusive lock on the data file right at the start and releasing it right at the end. As far as I can tell, this eliminates the error @ezbik was experiencing, but it's always possible I broke something else subtle.

Also, there is a branch that only happens if we're not on Solaris...I certainly hope no-one is.

### DIFF
--- a/rss2email/feeds.py
+++ b/rss2email/feeds.py
@@ -230,7 +230,7 @@ class Feeds (list):
                 return datafile
         return datafiles[0]
 
-    def load(self, lock=True, require=False):
+    def load(self, require=False):
         _LOG.debug('load feed configuration from {}'.format(self.configfiles))
         if self.configfiles:
             self.read_configfiles = self.config.read(self.configfiles)
@@ -238,9 +238,9 @@ class Feeds (list):
             self.read_configfiles = []
         _LOG.debug('loaded configuration from {}'.format(
                 self.read_configfiles))
-        self._load_feeds(lock=lock, require=require)
+        self._load_feeds(require=require)
 
-    def _load_feeds(self, lock, require):
+    def _load_feeds(self, require):
         _LOG.debug('load feed data from {}'.format(self.datafile_path))
         if not _os.path.exists(self.datafile_path):
             if require:
@@ -258,10 +258,7 @@ class Feeds (list):
         except IOError as e:
             raise _error.DataFileError(feeds=self) from e
 
-        locktype = 0
-        if lock:
-            locktype = _fcntl.LOCK_SH
-            _fcntl.lockf(self.datafile.fileno(), locktype)
+        _fcntl.lockf(self.datafile, _fcntl.LOCK_SH)
 
         self.clear()
 
@@ -288,10 +285,6 @@ class Feeds (list):
         _LOG.setLevel(level)
         _LOG.handlers = handlers
         self.extend(feeds)
-
-        if locktype == 0:
-            self.datafile.close()
-            self.datafile = None
 
         for feed in self:
             feed.load_from_config(self.config)
--- a/rss2email/feeds.py
+++ b/rss2email/feeds.py
@@ -132,18 +132,18 @@ class Feeds (list):
     datafile_version = 2
     datafile_encoding = 'utf-8'
 
-    def __init__(self, configfiles=None, datafile=None, config=None):
+    def __init__(self, configfiles=None, datafile_path=None, config=None):
         super(Feeds, self).__init__()
         if configfiles is None:
             configfiles = self._get_configfiles()
         self.configfiles = configfiles
-        if datafile is None:
-            datafile = self._get_datafile()
-        self.datafile = _os.path.realpath(datafile)
+        if datafile_path is None:
+            datafile_path = self._get_datafile_path()
+        self.datafile_path = _os.path.realpath(datafile_path)
         if config is None:
             config = _config.CONFIG
         self.config = config
-        self._datafile_lock = None
+        self._datafile = None
 
     def __getitem__(self, key):
         for feed in self:
@@ -214,7 +214,7 @@ class Feeds (list):
                 [_os.path.join(config_dir, 'rss2email.cfg')
                  for config_dir in config_dirs]))
 
-    def _get_datafile(self):
+    def _get_datafile_path(self):
         """Get the data file path
 
         Following the XDG Base Directory Specification.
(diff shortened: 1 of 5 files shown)


rss2email/feeds.py
```diff
@@ -242,37 +265,19 @@ def load(self, lock=True, require=False):
 
     def _load_feeds(self, lock, require):
```
Ekleog: I would guess the `lock` argument is no longer required

rss2email/feeds.py
```diff
@@ -337,32 +339,37 @@ def save(self):
...
             _os.makedirs(dirname, mode=0o700, exist_ok=True)
-        tmpfile = dst_config_file + '.tmp'
-        with open(tmpfile, 'w') as f:
+
+        with open(dst_config_file, 'w') as f:
+            _fcntl.lockf(f, _fcntl.LOCK_EX) # lock for writing
             self.config.write(f)
             f.flush()
             _os.fsync(f.fileno())
-        _os.replace(tmpfile, dst_config_file)
```
Ekleog: i think that for the config file, the write-then-mv behavior should be preferred, as `mv` is (usually) atomic, while write is not -- this reduces the risk of losing the configuration. Plus, if multiple processes are attempting configuration change at the same time, something is clearly wrong, and if there is a race condition it's not that bad.

rss2email/feeds.py
```diff
@@ -337,32 +339,37 @@ def save(self):
...
+        # of the feeds object, so the following is safe.
+
+        # We don't want to close the file and release the lock (allowing
+        # another rss2email to overwrite the data under us), so we do a
+        # seek and truncate
+        self.datafile.seek(0)
+        self._save_feed_states(feeds=self, stream=self.datafile)
+        self.datafile.truncate()
+        self.datafile.flush()
+        _os.fsync(self.datafile.fileno())
```
Ekleog: Suggestion: for the same reason as for the config, either add a `# TODO figure out a way to make this write atomic` or figure the thing out now, by doing something like write to a tmp file, locking the tmp file, mv-ing it in place and then closing the now-deleted file -- I'm not sure that's proper posix, though.

Ekleog:
(Hmm, looks like that went out as “requested changes”… I meant it as “r+ from me once the below is handled”, sorry!)

Profpatsch:
まったく. We should really return to sqlite for the config file.

rss2email/feeds.py
kaashif: Yeah. And actually, it was always called with lock=True anway so it was never required.

rss2email/feeds.py
kaashif: Yeah, if r2e dies in the middle, bad things will happen, what I am doing is not at all atomic. The only way out, I suppose, is to have a dedicated lock file which we acquire an exclusive lock on right at the start, then we do the atomic copy to temp file, modify, and rename.

kaashif:
Do you think it'd be a good idea to explicitly have a lockfile that we lock?

Then all of the file writing stuff won't have to even care about locking since we can just assume we're the only one running - I think it'll be a lot clearer.

Ekleog:
That feels right to me!

kaashif:
Great, I'll implement that this weekend.

kaashif:
I think this is a pretty simple solution - we use a lock file to ensure only one instance runs per user. If you run a lot of instances in parallel, you see (in the debug logs) that the next instance only acquires the lock after the previous instance finishes.

I also removed all of the "lock" options, since we always lock. So this solution is simpler, better, and doesn't break anything.

Profpatsch:
Can we have a test that tests this somehow?

kaashif:
Yeah, I'll write one. I think maybe the best way to test it would be to run some processes in parallel, make sure each "lock acquired" log message only appears after another process has written the data file, and also make sure all processes terminate successfully (before this change, some fail due to files not existing).

kaashif:
While writing the test for this, I eliminated the need to specify PYTHONPATH (the test script adds .. to sys.path itself) and PATH (the test script always uses the r2e script located in the directory above the directory with the test script). I was trying to diagnose and fix the test failures due to this error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/build/rss2email/test/../r2e'
```

But this file clearly should exist, right? I don't understand what's happening, how is Travis running the tests such that `rss2email/r2e` doesn't exist? For what it's worth, the test passes on my machine.

Profpatsch:
what’s the status of this? Can we rebase on master, I think we changed the logic to how the tests add PYTHONPATH

kaashif:
I'll try to get these tests working this weekend, but I have no idea what's wrong, surely the file `/build/rss2email/test/../r2e` does exist. I really want to see that green tick, so I'll investigate. Thanks for the reminder.

kaashif:
There, I fixed the tests! Things to note:

* The test.py script uses the r2e script at the path `R2E_PATH` (absolute path to script file) or `test/../r2e` if that's not available. The Nix test command passes this environment variable. The problem before was that I thought the Nix environment the test script ran in looked like the repo, apparently it doesn't, there is no file `test/../r2e`.

* The `directory` keyword argument was added in Python 3.7, so the unit tests were broken pre-3.7, I fixed this (without using os.chdir...).

* The tests are parametrized without adding any dependencies. If you look at the CI output, you'll notice that each config file run is now listed as a separate test.

I'm pretty sure this is good to be merged now.

One comment (that doesn't block this PR from being merged): why do we run the CI unit tests in Nix? This means that if there is a Nix related break in the tests, potential contributors will have to learn something extra to debug it, causing frustration (e.g. in #97). I think we should just use the [Travis way of running test for multiple Python versions](https://blog.travis-ci.com/2019-08-07-extensive-python-testing-on-travis-ci) and keep Nix for local development. That way contributors would just have to know what pip and setup.py do, which they hopefully already do.

kaashif:
I want to merge this soon, the test improvements in particular. Any objections? @Ekleog @Profpatsch?

Ekleog:
Looks good to me! There's just the removal of `test_send` which I kind of wonder why, but apart from that there's not much for me to say :)

kaashif:
Thanks! test_send was split up, not removed exactly, but it's not obvious from the diff - all the same tests are run. I'll merge this later, after making sure the merge will work correctly.

PR #95, opened 2020-02-05
Add --maximum option to run command, fix bugs in new user agent configurability

jikamens:
The first commit in this PR adds a new option to the `run` command, `--maximum`, which specifies the maximum number of new entries to process for each feed. This is useful for people who want to avoid getting big bursts of email from feeds that tend to add many entries at once, and also useful during development and debugging so you can use a real feed during testing without getting a ton of emails every time you test.

The rest of the commits are various fixes to the functionality added in release 3.11 to allow the user agent string to be configured. See the descriptions of the individual commits for details.

### DIFF
--- a/r2e.1
+++ b/r2e.1
@@ -67,7 +67,7 @@ optional \fI<email>\fR argument is the email address to send new items
 to, overriding the default address for this particular feed.  Repeat
 for each feed you want to subscribe to.
 .TP
-.B run \fR[\fI\-\-no-send\fR] \fR[\fI<index>\fR [\fI<index>\fR ...]]
+.B run \fR[\fI\-\-no-send\fR] \fR[\fI\-\-maximum <num>\fR] \fR[\fI<index>\fR [\fI<index>\fR ...]]
 Scan the feeds and send emails for new items. This can be run in a cron
 job.
 .P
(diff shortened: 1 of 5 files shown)


Profpatsch:
Hi @jikamens, thank you for your contribution!

I think we can merge the fixes directly, could you split out the fixes in a separate PR?

As for the new option, I would like to consider some alternatives in the design space. It feels very specific to one use-case and can maybe be achieved trough some more general option. Then again, maybe it’s a useful enough feature to merit a specific command line option.

jikamens:
User agent fixes pulled into separate PR at https://github.com/rss2email/rss2email/pull/97.

jikamens:
Does, "I would like to consider some alternatives in the design space," mean that you are looking for such consideration / input from _me_, or is that a statement that this is something you and the other maintainers of the package want to bat around a bit?

Because if the former, then my personal opinion is squarely in the "it's a useful enough feature to merit a specific command line option" camp, so I'm not sure my input would be valuable.

Though I suppose I'd entertain the argument that it could be a configuration setting that can be overridden on a per-feed basis rather than a command-line argument. :man_shrugging:

Ekleog:
One alternative design might be to extend the post-process hooks, so that they can decide to drop mail, and get the information about the number of the email that's being sent in this run, for instance.

jikamens:
>One alternative design might be to extend the post-process hooks, so that they can decide to drop mail, and get the information about the number of the email that's being sent in this run, for instance.

That's definitely not the goal of my change. I don't want to _drop_ emails, I want to _delay_ them. My change just stops processing each feed after the specified number of entries, leaving the remaining entries to be processed in a later run. Your change would process all the entries and then throw away some of them; that's not my intent.

Ekleog:
Well, we could add yet another option to skip the emails and re-handle them later on if they're still in the rss feed by then -- but this feels like a really niche use case to me. TBH, I didn't even think it'd be what you were advocating, sorry for the confusion!

jikamens:
Closing because apparently this change is not going to be merged. :man_shrugging:

PR #97, opened 2020-02-06
Fixes to the configurable user-agent changes introduced in 3.11

jikamens:

PR #100, opened 2020-02-16
Fixes to the configurable user-agent changes introduced in 3.11

kaashif:
This PR is the first few commits of #97, plus two tests that:

* The substituted values aren't written to the config file when you add a new feed

* Old badly substituted per-feed configs are corrected (this only happens if you added a feed in v3.11 and had the default user agent)

I think there's still some other stuff in #97 that would be good to pick up, mostly test improvements, but I think these bug fixes should be merged first. I'll make a PR later on to improve tests, I've done a bit of that in #94, but smaller PRs are easier to review (and take less time to make).

cc @jikamens

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -168,6 +168,9 @@ class Feed (object):
     _default_configured_attributes[
         _default_configured_attributes.index('from')
         ] = 'from_email'  # `from` is a Python keyword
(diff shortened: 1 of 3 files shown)


jikamens:
Thank you for helping get these fixes across the finish line. I appreciate it.

jikamens:
Ping?

PR #102, opened 2020-03-07
Include feed info in error message when fetch times out

jikamens:
Before this change, when a feed times out, this is the error the user sees:

```
2020-03-07 10:01:02,784 [ERROR] 60 second timeout exceeded
```

This is unhelpful.

This commit modifies the behavior so that the feed information appears in the error:

```
2020-03-07 10:05:05,697 [ERROR] [feed-name] ([url] -> [email]): 60 second timeout exceeded
```

### DIFF
--- a/rss2email/error.py
+++ b/rss2email/error.py
@@ -43,13 +43,15 @@ class RSS2EmailError (Exception):
 class TimeoutError (RSS2EmailError):
     def __init__(self, time_limited_function, message=None):
         if message is None:
-            if time_limited_function.error is not None:
-                message = (
-                    'error while running time limited function: {}'.format(
-                        time_limited_function.error[1]))
-            else:
-                message = '{} second timeout exceeded'.format(
-                    time_limited_function.timeout)
+            message = ''
+        else:
+            message += ': '
+        if time_limited_function.error is not None:
+            message += 'error while running time limited function: {}'.format(
+                time_limited_function.error[1])
+        else:
+            message += '{} second timeout exceeded'.format(
+                time_limited_function.timeout)
         super(TimeoutError, self).__init__(message=message)
         self.time_limited_function = time_limited_function
 
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -356,7 +356,7 @@ class Feed (object):
         kwargs = {}
         if proxy:
             kwargs['handlers'] = [_urllib_request.ProxyHandler({'http':proxy})]
-        f = _util.TimeLimitedFunction(timeout, _feedparser.parse)
+        f = _util.TimeLimitedFunction(timeout, _feedparser.parse, message=str(self))
         return f(self.url, self.etag, modified=self.modified, **kwargs)
 
     def _process(self, parsed):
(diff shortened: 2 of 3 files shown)


rss2email/error.py
```diff
@@ -43,13 +43,15 @@ def log(self):
 class TimeoutError (RSS2EmailError):
     def __init__(self, time_limited_function, message=None):
```
Profpatsch: `message=""`

then the check below can be left out

rss2email/error.py
```diff
@@ -43,13 +43,15 @@ def log(self):
...
-                    time_limited_function.timeout)
+            message = ''
+        else:
+            message += ': '
+        if time_limited_function.error is not None:
+            message += 'error while running time limited function: {}'.format(
+                time_limited_function.error[1])
+        else:
+            message += '{} second timeout exceeded'.format(
+                time_limited_function.timeout)
```
Profpatsch: I’m not convinced pulling the logic out of the `message is None` check will keep the intended semantics between the `time_limited_function` and `message` arguments.

Previously, it passed `message` through as-is if it was not `None`, now it always adds stuff.

rss2email/error.py
jikamens: The code as I've written it is correct. It would not be correct to append `': '` to `message` if it's an empty string.

rss2email/error.py
jikamens: There is as far as I can tell only one place this is used in the entire code-base and I believe my change is correct for that use-case.

Ekleog:
@jikamens It looks like there's significant overlap with https://github.com/rss2email/rss2email/pull/133, which I merged before seeing your PR. Am I correct if I think that there is no longer a need to merge this? (sorry for not having noticed your earlier PR before merging #133)

jikamens:
> @jikamens It looks like there's significant overlap with #133, which I merged before seeing your PR. Am I correct if I think that there is no longer a need to merge this? (sorry for not having noticed your earlier PR before merging #133)

I don't know. I've given up on contributing to this project because I find it too difficult.

auouymous:
They do the same thing, sorry for not seeing this sooner.

I also missed the message parameter on TimeoutError. TimeLimitedFunction is the only code that uses TimeoutError and it doesn't pass a message. It is also tightly coupled to TimeLimitedFunction due to its time_limited_function parameter. Should the message parameter be removed? Or should it be prefixed with time_limited_function.name in case it is ever used by future code?

Ekleog:
@jikamens This is sad :( sorry to hear that! unfortunately we kind of just catched the dropped ball with maintenance of this project (which AFAIU was already a dropped ball, all caused by moving priorities), so it's not easy to maintain the project, hence the significant delays. Anyway, sorry about that! And hope you find other OSS projects to contribute to :)

@auouymous I'm not sure what you missed? IMO not adding the feed info when the message is explicitly provided makes sense. But seeing as, as you say, the `message` parameter is never used, it makes sense to remove it, and I'll be happy to accept a PR doing so :)

Closing this, as it has been landed in a different way and doesn't appear to have a clear way forward — feel free to yell at me if I'm wrong

Issue #11, opened 2017-12-05
SHA1 of entry content as internal id is not reliable


AndreyMZ:
#### Problem description

When
```
trust-guid = False
trust-link = False
```
the SHA1 of entry content is used as internal id.
But this is not reliable because of https://github.com/kurtmckee/feedparser/issues/100

#### Impact

With such configuration for some feeds (e.g. [tar](https://savannah.gnu.org/news/atom.php?group=tar)) rss2email will send similar email messages (with no visible changes) on every run.

Ekleog:
@AndreyMZ Do you have any idea of what could be done to alleviate this issue?

AndreyMZ:
I think, the best solution is to fix the feedparser issue #100. Until that I can only suggest not to use the affected configuration.

Ekleog:
Cool, I've relabeled to mention that, thank you for your feedback!

PR #103, opened 2020-04-02
fix html2text config options

auouymous:
This uses the correct path for the html2text config options.

### DIFF
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -59,11 +59,13 @@ class Config (_configparser.ConfigParser):
         """
         if section not in self:
             section = 'DEFAULT'
-        _html2text.UNICODE_SNOB = self.getboolean(
+        _html2text.config.UNICODE_SNOB = self.getboolean(
             section, 'unicode-snob')
-        _html2text.LINKS_EACH_PARAGRAPH = self.getboolean(
+        _html2text.config.LINKS_EACH_PARAGRAPH = self.getboolean(
             section, 'links-after-each-paragraph')
-        _html2text.BODY_WIDTH = self.getint(section, 'body-width')
+        # hack to prevent breaking the default in every existing config file
+        body_width = self.getint(section, 'body-width')
+        _html2text.config.BODY_WIDTH = 0 if body_width < 0 else 78 if body_width == 0 else body_width
 
 
 CONFIG = Config()
@@ -190,7 +192,8 @@ CONFIG['DEFAULT'] = _collections.OrderedDict((
         ('unicode-snob', str(False)),
         # Put the links after each paragraph instead of at the end.
         ('links-after-each-paragraph', str(False)),
-        # Wrap long lines at position. 0 for no wrapping.
+        # Wrap long lines at position.
+        # Any negative value for no wrapping, 0 for 78 width (compatibility), or any positive width.
         ('body-width', str(0)),
 
         ### Mailing
--- a/r2e.1
+++ b/r2e.1
@@ -230,7 +230,8 @@ Use Unicode characters instead of their ascii psuedo-replacements
 .IP links-after-each-paragraph
 Put the links after each paragraph instead of at the end.
 .IP body-width
-Wrap long lines at position. 0 for no wrapping.
+Wrap long lines at position. Any negative value for no wrapping, 0 for 78 width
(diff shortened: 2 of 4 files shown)


rss2email/config.py
```diff
@@ -59,11 +59,11 @@ def setup_html2text(self, section='DEFAULT'):
         """
         if section not in self:
             section = 'DEFAULT'
-        _html2text.UNICODE_SNOB = self.getboolean(
+        _html2text.config.UNICODE_SNOB = self.getboolean(
             section, 'unicode-snob')
-        _html2text.LINKS_EACH_PARAGRAPH = self.getboolean(
+        _html2text.config.LINKS_EACH_PARAGRAPH = self.getboolean(
             section, 'links-after-each-paragraph')
-        _html2text.BODY_WIDTH = self.getint(section, 'body-width')
```
Profpatsch, line 66: Does that mean these lines had no effect before?

auouymous: Correct. I assume they worked (were tested) when added and a change in html2text broke them. I was unable to find such a change.

Profpatsch:
The golden tests are failing (just a whitespace diff), so I guess the options show some effect now.

auouymous:
I have a few issues on an unmodified copy of master.

```
ValueError: error processing disqus/1.config
-X-RSS-ID: ab03f2100069a1cd0876b997be87976c18d48e8a
+X-RSS-ID: a52375ec78a988241fe9864a2243d4d910538d52
```

And specifying a group of tests or a single as mentioned in test/README throws a `AttributeError: module '__main__' has no attribute 'test/allthingsrss'` error.

Am I the only one with those two errors?

test/allthingsrss/3.expected and test/allthingsrss/4.expected are clearly not wrapped to 70 and both set body-width=70. Wrapping those files will get them passing. The biggest problem is that html2text defaults to 78 and r2e assumes 0 when body-width is not given. I'm guessing r2e never had working wrapping because every test is wrapped to 78, but correctly setting body-width to 0 causes them all to fail. Should I unwrap the expected output of every test? Or should I add body-width=78 to all tests except the two that set it to 70?

Another problem is that r2e defaults to body-width=0 in the config and everyone probably has it at default since changing it has never had any effect. But this entire time, html2text has been wrapping to its default of 78. Anyone who upgrades to the release with this patch will find all new emails to be unwrapped. The easiest way to handle it is to force anyone who doesn't like the new unwrapped emails to go change body-width to 78. Or we can detect 0 and pass to html2text as 78, and change default to -1 and detect -1 and pass to html2text as 0. The 0/-1 detection means tests won't need to be modified to get them to pass, it also provides the smoothest upgrade for users. Thoughts?

kaashif:
Sorry about the README being wrong, I fixed that in my PR here: https://github.com/rss2email/rss2email/pull/94/files#diff-df487c01051d76eefe58f127d31cac6a but we haven't merged it yet. (In that PR I also change the test cases to be run separately as unittest test cases instead of a single test for all configs, so you can better see which tests are failing)

That test failure is a long-standing intermittently flaky test, due to the orders of attributes in the HTML changing between runs (see https://github.com/rss2email/rss2email/issues/11), changing the hash. Don't worry about that particular test failure. We might have to fix this soon though, contributors always ask about this.

If users don't change their configs, they should see no change in output, but that's not really possible in this case. Making people change their configs would be the easiest solution to write, but probably the most difficult solution in terms of hate mail haha. If we can't maintain perfect backwards compatibility (impossible since it's possible some user explicitly set body-width=1 or something stupid and expects 78 wrapped output), then I think making any value less than or equal to zero really mean 78 is the best option.

Kind of a hack, but I don't see any other way to do this.

auouymous:
Now it detects any negative value and uses 0, and detects 0 and uses 78. Should the html2text options be mentioned in the README now that they actually work?

kaashif:
There's a single line in the man page that needs to be updated: https://github.com/rss2email/rss2email/blob/master/r2e.1#L233. I think that would be enough documentation - the documentation in the README actually isn't complete and isn't intended to be, as far as I know, it just says "look at config.py". Once the man page is updated, this is good to be merged, I think.

Good work finding this bug!

auouymous:
Man page fixed.

PR #104, opened 2020-04-02
add encode-date-field option to disable encoding for MUA's that lack support

auouymous:
Not all MUA's support encoded date fields, Mutt doesn't and Thunderbird might not unless fixed since https://bugzilla.mozilla.org/show_bug.cgi?id=276199 decided it was an RFC violation.

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -45,12 +45,14 @@
 """Define the ``Feed`` class for handling a single feed
 """
 
+import calendar as _calendar
 import collections as _collections
 import platform
 from email.message import Message
 from email.mime.message import MIMEMessage as _MIMEMessage
 from email.mime.multipart import MIMEMultipart as _MIMEMultipart
 from email.utils import formataddr as _formataddr
+from email.utils import formatdate as _formatdate
 from email.utils import parseaddr as _parseaddr
 import hashlib as _hashlib
 import html.parser as _html_parser
@@ -609,7 +611,7 @@ class Feed (object):
                 if entry.get(kind, None):
                     datetime = entry[kind]
                     break
-        return _time.strftime("%a, %d %b %Y %H:%M:%S -0000", datetime)
+        return _formatdate(_calendar.timegm(datetime))
 
     def _get_entry_name(self, parsed, entry):
         """Get the best name


rss2email/email.py
```diff
@@ -146,8 +146,11 @@ def get_message(sender, recipient, subject, body, content_type,
         message.set_payload(body, charset=charset)
     if extra_headers:
         for key,value in extra_headers.items():
-            encoding = guess_encoding(value, encodings)
-            message[key] = _Header(value, encoding)
+            if config.getboolean(section, 'encode-date-field') or key != 'Date':
```
Profpatsch, line 149: That’s inverted, the test should be   key == Date  and not <config>

rss2email/email.py
```diff
@@ -146,8 +146,11 @@ def get_message(sender, recipient, subject, body, content_type,
...
+                encoding = guess_encoding(value, encodings)
```
Profpatsch, line 150: what exactly does “encoding the date field” mean? Can you give some more background?

Profpatsch: Also a FAQ entry would be much appreciated, this sounds like an issue other users might stumble into.

rss2email/email.py
auouymous: The logic is correct. Inverting it would require swapping the true and false branch bodies, is that what you want?

rss2email/email.py
auouymous: The _Header function QP (RFC 2047) encodes the header contents, and it does so for all headers, including Date.

Date: Mon, 30 Mar 2020 09:18:24 -0000
becomes
Date: =?utf-8?q?Mon=2C_30_Mar_2020_09=3A18=3A24_-0000?=

A locale might generate non-ASCII characters for the day, month, or timezone, and must be QP encoded, or risk being rejected by smtp servers. The Thunderbird issue questioned if encoding the date field violated both RFC 822 and the newer RFC 2822 which defines a syntax that explicitly uses english names for day, month and timezone. Any strict MUA would reject QP encoded date headers due the sender's violation of the RFC.

Another option is to force english in srtftime().

```
with setlocale('C'):
    return _time.strftime("%a, %d %b %Y %H:%M:%S -0000", datetime)
```

Then r2e wouldn't be violating the RFC and the encode-date-field option wouldn't be needed. The 'Date' field check to bypass _Header() would still be needed. I am still investigating why, but it would be nice if QP encoding wasn't always enabled for all headers. Printing _Header() output doesn't QP encode it, even if it conains utf-8 characters, something later on is forcing it to always encode. If that was fixed, it would not only reduce the size of outgoing emails but checking for 'Date' field wouldn't be needed.

rss2email/email.py
Profpatsch: Well, the logic is correct but it’s completely unreadable.

I’d rather have one more branch for the Date case and mention `guess_encoding` twice.

```
if == Date:
  if config_set: guess_encoding
  else: dont
else:
  guess_encoding
```

rss2email/email.py
Profpatsch: If we can solve it generally without undue backwards incompatible changes, I’d be in favor of that. If that’s not viable, adding an option and an troubleshooting entry is a good next-best solution.

auouymous:
Sorry for the long delay. My headers were QP encoded because I put "UTF-8" first in encodings to fix an issue back in 2.71. This new patch never QP encodes the Date header and always forces a 'C' locale so the formatted date is always in english and compliant with the RFC.

rss2email/feed.py
```diff
@@ -562,7 +565,17 @@ def _get_entry_date(self, entry):
...
+        @_contextlib.contextmanager
+        def setlocale(name):
+            with LOCALE_LOCK:
+                saved = _locale.setlocale(_locale.LC_ALL)
+                try:
+                    yield _locale.setlocale(_locale.LC_ALL, name)
+                finally:
+                    _locale.setlocale(_locale.LC_ALL, saved)
+        with setlocale("C"):
+            return _time.strftime("%a, %d %b %Y %H:%M:%S -0000", datetime)
```
amiryal, line 578: After a bit of research, I found [email.utils.formatdate](https://docs.python.org/3/library/email.utils.html#email.utils.formatdate), which gives you a compliant header value with much less effort, so this line would simply change to:
```python
…
from email.utils import formatdate as _formatdate
…
        …
        _formatdate(_time.mktime(datetime))
…
```

rss2email/email.py
```diff
@@ -146,8 +146,11 @@ def get_message(sender, recipient, subject, body, content_type,
         message.set_payload(body, charset=charset)
     if extra_headers:
         for key,value in extra_headers.items():
-            encoding = guess_encoding(value, encodings)
-            message[key] = _Header(value, encoding)
+            if key == 'Date':
+                message[key] = value
+            else:
+                encoding = guess_encoding(value, encodings)
+                message[key] = _Header(value, encoding)
```
amiryal, line 153: This would be redundant with [one of my changes](https://github.com/rss2email/rss2email/pull/166/commits/db959a8e1975211828806d16ac75513a37dcf9c6#diff-5858faa28f9fb55ca9746766095d02da651d8a85177f7b3d1c20d405a1d28c54R140) in #166.

amiryal (changes requested):

amiryal:
Incidentally, by running the test suite with [my suggested fix](https://github.com/rss2email/rss2email/pull/104#discussion_r596803787) applied, I have exposed a bug in this line:
```python
return _time.strftime("%a, %d %b %Y %H:%M:%S -0000", datetime)
```
Since it uses `-0000` instead of `%z`, the timestamps that it produces are all wrong, except for values that happen to be in GMT-0 by chance. More reason to accept my suggested changes and merge this PR.

auouymous:
The problem is that `datetime` is UTC, `mktime` converts it to localtime and `formatdate` wants UTC. I replaced `mktime` with `calendar.timegm` to prevent conversion to UTC.

amiryal:
> The problem is that `datetime` is UTC, `mktime` converts it to localtime and `formatdate` wants UTC. I replaced `mktime` with `calendar.timegm` to prevent conversion to UTC.

Correct! I have edited my earlier comment to avoid confusion. Seems like the standard library needs to be organised better.

amiryal (approved):

Ekleog:
Looks good to me, thank you!

Issue #96, opened 2020-02-05
URLs line-wrapped and therefore broken when using text format emails


jikamens:
When emails are being sent in text format instead of HTML, links are converted to markdown syntax and then the body of the text is line-wrapped. The line-wrapping doesn't treat links in any special way, which means that links with hyphens in them can get wrapped at the hyphen, thus breaking the link.

f6k:
Hello. I have the same issue. I digged a little and this is what I found.

Actually the thing is that `html2markdown` (aka `html2text`) (version 2018.1.9 here) have defaulst to put links inline in the document *and* to wrap links. This can be seen in its `config.py`. On my box, Debian, it's located in `/usr/lib/python3/dist-packages/html2text/config.py` where we can read that :

    # Use inline, rather than reference, formatting for images and links
    INLINE_LINKS = True
    (...)
    WRAP_LINKS = True

If we jump to the `config.py` of `rss2email` we see that nothing is declared for those variables. The funny thing is that there is actually something for putting the links after each paragraph, but since, by default, `html2markdown` put links inline, this one is obviously ineffective. My workaround would be to add two lines in config.py of `rss2email` under the `def setup_html2text(self, section ='DEFAULT')` about `INLINE_LINKS` and `WRAP_LINKS`:

    (...)
    _html2text.LINKS_EACH_PARAGRAPH = self.getboolean(section, 'links-after-each-paragraph')
    _html2text.BODY_WIDTH = self.getint(section, 'body-width')
    _html2text.INLINE_LINKS = self.getboolean(section, 'inline-links')
    _html2text.WRAP_LINKS = self.getboolean(section, 'wrap-links')

And then, in `~/.config/rss2email.cfg` having something like in `[DEFAULT]`:

    inline-links = False
    links-after-each-paragraph = True
    wrap-links = False

So, as a result, we should NOT have inline links but unwrapped links put at references at the end of each paragraph. One could play with `inline-links` and `wrap-links` to have the links inline in the document *and* not wrapped.

**This has not been properly tested yet**. I'll make an update if I find more.

By the way, while I'm at it. At first, even if the links were inline, I thinked that I could just put `body-width = 0` so the text (and the then the links inline) should not be wrapped. But it doesn't work. I'm still searching why `rss2email` doesn't properly take my modification for `body-width`. Whatever I put in my `rss2email.cfg` it sticks to the default declared by `html2markdown` (`BODY_WIDTH = 78`).

Profpatsch:
[...]

PR #106, opened 2020-04-17
Add inline-links option

alyssais:
I find the default output very difficult to read, because
of (potentially very long) URIs showing up in the middle of the text.

### DIFF
--- a/r2e.1
+++ b/r2e.1
@@ -229,6 +229,8 @@ Optional CSS styling
 Use Unicode characters instead of their ascii psuedo-replacements
 .IP links-after-each-paragraph
 Put the links after each paragraph instead of at the end.
+.IP inline-links
+Use inline, rather than reference, formatting for images and links.
 .IP body-width
 Wrap long lines at position. Any negative value for no wrapping, 0 for 78 width
 (compatibility), or any positive width.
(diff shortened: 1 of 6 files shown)


auouymous:
I agree. I've been using an inline-links and wrap-links patch.

Any chance you could also add the wrap-links option? Not wrapping inline-links prevents the entire paragraph from being wrapped, but it becomes useful when inline-links is false, because wrapped links can be harder to select.

It would also fully solve https://github.com/rss2email/rss2email/issues/96.

kaashif:
Thanks for this contribution, this would be a useful config option to have. One comment: instead of modifying the existing test 3.config and 3.expected, can you copy them and make the changes to a new test 5.config and 5.expected? The changes to the test output are fairly significant.

Other than that I think this is good to be merged if there are no objections from other maintainers.

Ekleog:
@alyssais I've allowed myself to implement @kaashif 's suggestion myself, as it was a trivial-ish change (two `cp` and one `git checkout`). So this is now ready to merge, thank you for the PR!

Ekleog:
Oh. Well, and as we all forgot about the changelog, I'll just update it in a follow-up commit :)

PR #109, opened 2020-04-19
Make _get_entry_content() more deterministic

martin-v:
If a content exist don't use the summary as content, as this may cause the content exist but is ignored.
But add the summary to the content if both exist.


Ekleog:
Looks like the tests failed — I'm not sure I understand what your PR is exactly about, but maybe if you adjusted the tests so that they don't fail it'd help showcase what changes you think would be beneficial?

Ekleog:
@martin-v Are you still planning to come back to this PR?

martin-v:
I have looked at this change again and I think it is better to leave it as it is.

PR #112, opened 2020-05-22
New "reply-changes" option and some fixes

AndreyMZ:
### Changes

New features:
 * https://github.com/rss2email/rss2email/commit/9d7cb815a212c607f3b2a5e33bbabfe3f892000a Add "reply-changes" option.

Changes in existing features:
 * https://github.com/rss2email/rss2email/commit/18190d5e289440cd0de6d3362c2d27521b160687 Use `platform.node()` instead of "dev.null.invalid" in "Message-ID" header.
 * https://github.com/rss2email/rss2email/commit/b122ec8d70b2f1301860ab7ec0a748e4e1defd69 Drop support for Python 3.4. Add support for Python 3.8.
   Note: Python 3.4 has reached end-of-life on 2019-03-18 (see the [Development Cycle](https://devguide.python.org/devcycle/#end-of-life-branches)).
 
Bug fixes:
 * https://github.com/rss2email/rss2email/commit/a8401ac5b651489792ba0ee1e1f5c3696958a0c6 Revert "Do not care about SunOS locking". This reverts commit a418a2d.
   https://github.com/rss2email/rss2email/commit/0e56a9a7140d4afb6a32f2fa0f3e365d46be69ef Do not care about SunOS locking, but do care about Windows.
   Note: @kaashif, please be careful next time.

The rest are bug fixes and refactoring mostly in tests. I did not want, but I had to do this =).

### Tested on

- Debian 10.4 + Python 3.8.3  ([python:3.8.3-buster](https://github.com/docker-library/python/blob/f59139d685b3b39d4452a686521e03fa093efa25/3.8/buster/Dockerfile))
- Debian 10.4 + Python 3.7.7  ([python:3.7.7-buster](https://github.com/docker-library/python/blob/43e801fc6ced5379e70a058a3d4fa3c2ee8c8976/3.7/buster/Dockerfile))
- Debian 10.4 + Python 3.6.10 ([python:3.6.10-buster](https://github.com/docker-library/python/blob/83822b9a8bc8d7f146c194b595702f4681fb7500/3.6/buster/Dockerfile))
- Debian 10.4 + Python 3.5.9  ([python:3.5.9-buster](https://github.com/docker-library/python/blob/f4eaaa988af2b8a56bc2ca1e1a69209217ccbbc2/3.5/buster/Dockerfile))
- CentOS 6.10 + Python 3.6.8 (`python36u` from [IUS](https://ius.io/))
- Windows 10 + Python 3.8.3
- Windows 10 + Python 3.7.7
- Windows 10 + Python 3.6.8

### Possible further improvements

Run all test modules (not only `./test/test.py`) in [nix/release.nix:29-31](https://github.com/rss2email/rss2email/blob/master/nix/release.nix#L29-L31). I am not familiar with this CI, so let somebody other do this.


AndreyMZ:
I do not understand this errors.
~~~
  File "/nix/store/8p3zlx3p2xv4i2zysm1kh822kcnx791m-rss2email-python_3_8/bin/r2e", line 2
    export PATH='/nix/store/1vjqqnc3gwr8gbrhvlg298ysi69gwl4v-python3-3.8.0/bin:/nix/store/8p3zlx3p2xv4i2zysm1kh822kcnx791m-rss2email-python_3_8/bin:/nix/store/1hnakfyg0w56374dyc5w2kq6wj0lmvjc-python3.8-html2text-2018.1.9/bin'${PATH:+':'}$PATH
           ^
SyntaxError: invalid syntax
~~~

kaashif:
Sorry about the regression, thanks for all the fixes and testing, I haven't been testing on Windows (as is obvious).

Since I now realise Windows is a use case we have to support, I think it would be a good idea to run the CI build on Windows (and macOS if we can get that for free), I think [GitHub actions supports this](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#specifying-a-python-version), I made an issue: #117. Travis doesn't support Python for Windows or macOS.

I also don't understand the error from CI, I think it's something nix is doing. I already had to put some hacks in to work around something nix was doing, I'm starting to think it's maybe not worth it. GitHub Actions is the way to go because of Windows support, I think.

I'll look at this PR when I get some time.

test/test.py
```diff
@@ -8,29 +8,28 @@
...
-# By default, we run the r2e in the dir above this script, not the
-# system-wide installed version, which you probably don't mean to
-# test. You can also pass in an alternate location in the R2E_PATH
-# environment variable.
-r2e_path = _os.getenv("R2E_PATH", _os.path.join(test_dir, "..", "r2e"))
+# Directory containing test feed data/configs
+test_dir = str(Path(__file__).absolute().parent.joinpath("data"))
 
 # Ensure we import the local (not system-wide) rss2email module
 sys.path.insert(0, _os.path.dirname(r2e_path))
```
Ekleog, line 28: With this change, the import path, that was before
`_os.getenv("R2E_PATH", _os.path.join(_os.path.dirname(_os.path.abspath(__file__)), "..", "r2e"))`
, is now
`os.getenv("R2E_PATH", str(Path(__file__).absolute().parent.parent.parent.joinpath("r2e")))`
, which is one `.parent` more than before.

I haven't investigated much further, but would bet this is the source of the issue that happens inside nix.

If I understand correctly, the meaning of the `R2E_PATH` has changed compared to what it was before. Before it was pointing to the executable. Now it is pointing to… the library? I'm far from being sure about it, as I haven't reviewed in enough depth.

If it's just changing the meaning of the R2E_PATH to point to the library instead, then it'd be possible to simply adjust the `R2E_PATH` variable in `nix/release.nix` to make things work again.

However… Seeing as it's no longer an executable, but the python library, referring to its location by its relative path to the current file should just work, and the whole `R2E_PATH` (which was a hack designed to be able to call the binary from the tests) could be removed altogether. Which should hopefully solve the issue :) (you can check that by just removing the `R2E_PATH` assignment in `nix/release.nix`, and the `os.getenv` will then follow the same path on nix as on non-nix)

cc @Profpatsch just in case, as I think you probably introduced R2E_PATH?

kaashif: Actually, I was the one who introduced R2E_PATH. R2E_PATH should be unnecessary, since the test script should be able to find the `r2e` binary from the `test/` directory using `../r2e`, but this wasn't working for some reason.

rss2email/feed.py
```diff
@@ -513,31 +534,43 @@ def _process_entry(self, parsed, entry):
...
-                # Newer versions of feedparser could return a dictionary
-                if isinstance(entry.id, dict):
-                    return entry.id.values()[0]
-                return entry.id
+            uid = self._get_entry_id(entry)
+            if uid:
+                return uid
+        return self._get_entry_hash(entry)
+
+    @lru_cache(1)
```
kaashif, line 552: This is fine, but I'm wondering why you put this here. Is it the case that we're calling `_get_entry_hash` on the same entry repeatedly, so that an LRU cache of size 1 is useful? I haven't examined the usages of this function in detail so that might very well be the case, I'm just curious.

rss2email/feeds.py
```diff
@@ -43,7 +43,11 @@
 from . import error as _error
 from . import feed as _feed
 
-import fcntl as _fcntl
+try:
+    import fcntl as _fcntl
+    UNIX = True
+except ImportError:
```
kaashif, line 49: Sorry about breaking this for Windows in the past. I suppose the correct thing for me to do would be to have a cross-platform lock object here, so we don't have to do this import check.

rss2email/main.py
```diff
@@ -163,25 +163,32 @@ def run(*args, **kwargs):
...
+        if args.func != _command.new:
+            feeds.load()
+        args.func(feeds=feeds, args=args)
     except _error.RSS2EmailError as e:
         e.log()
         if _logging.ERROR - 10 * args.verbose < _logging.DEBUG:
             raise  # don't mask the traceback
         _sys.exit(1)
+    finally:
+        if lockfile is not None:
```
kaashif, line 190: It is a bit of a shame that we have to do a try-finally instead of a `with`, but I guess this is needed to do nothing on Windows.

test/README
```diff
@@ -1,6 +1,7 @@
 Run tests with::
 
-  $ ./test/test.py
+  $ cd ./test
+  $ python -m unittest
```
kaashif, line 4: Splitting up the tests a bit would be good, I agree with this change.

test/README
```diff
@@ -10,19 +11,20 @@ saved (not mailed) and compared with the expected output for that
 feed/config pair.  For example, with a directory structure like::
 
   test
+  |-- data
```
kaashif, line 14: Why did you move all of the test configs and feed data into this `data/` directory? Is it something to do with the unittest test discovery maybe?

test/test.py
```diff
@@ -146,45 +145,42 @@ def run_single_test(self, dirname=None, config_path=None, force=False):
...
+            try:
+                _os.chdir(test_dir)
+                return super().translate_path(path)
+            finally:
+                _os.chdir(cwd)
 
     def log_message(self, format, *args):
         return
 
+def webserver_for_test_fetch(queue, num_requests, wait_time):
```
kaashif, line 166: Is there some reason you moved this out of the test? Since you had to add `for_test_fetch` to the function name, this seems to indicate to me that it really belongs inside `TestFetch`.

test/test.py
```diff
@@ -277,25 +258,25 @@ def test_fetch_parallel(self):
                     previous_line = line
             self.assertTrue(finish_precedes_acquire)
 
+def webserver_for_test_send(queue):
```
kaashif, line 261: Same comment as for `webserver_for_test_fetch`: why was this moved?

test/test.py
AndreyMZ: The value of `r2e_path` is the same. There is one `.parent` more, because the new `__file__` has one nested directory more than the previous one: `test/util/execcontext.py` vs `test/test.py`.

rss2email/feed.py
AndreyMZ: Yes, it is just such a case. The `_get_entry_hash` method may be called twice:
1. [`_process_entry`](https://github.com/rss2email/rss2email/blob/9d7cb815a212c607f3b2a5e33bbabfe3f892000a/rss2email/feed.py#L472) > [`_get_uid_for_entry`](https://github.com/rss2email/rss2email/blob/9d7cb815a212c607f3b2a5e33bbabfe3f892000a/rss2email/feed.py#L550) > `_get_entry_hash`
2. [`_process_entry`](https://github.com/rss2email/rss2email/blob/9d7cb815a212c607f3b2a5e33bbabfe3f892000a/rss2email/feed.py#L473) > `_get_entry_hash`

Also the `_get_entry_content` method may be called several times:
1. [`_get_entry_hash`](https://github.com/rss2email/rss2email/blob/9d7cb815a212c607f3b2a5e33bbabfe3f892000a/rss2email/feed.py#L554) > `_get_entry_content`
2. [`_get_entry_title`](https://github.com/rss2email/rss2email/blob/9d7cb815a212c607f3b2a5e33bbabfe3f892000a/rss2email/feed.py#L582) > `_get_entry_content`
3. [`_process_entry`](https://github.com/rss2email/rss2email/blob/9d7cb815a212c607f3b2a5e33bbabfe3f892000a/rss2email/feed.py#L521) > `_get_entry_content`

rss2email/feeds.py
AndreyMZ: That’s OK, it happens =).

rss2email/main.py
AndreyMZ: I agree. We may introduce a helper context manager, e.g. `Lock`, and use `with Lock()` instead of `try`/`finally`. But I think this is out of scope of this commit.

test/README
AndreyMZ: Nothing special. I just wanted not to mix code with data.

test/test.py
AndreyMZ: Yes, there is a reason. I encountered an error on running the test. I wrote it in the [commit](https://github.com/rss2email/rss2email/pull/112/commits/5b588dfc4b6df06ebdea48baa8e9367ac68a6b57) message:
~~~
AttributeError: Can't pickle local object 'TestFetch.test_delay.<locals>.webserver'
~~~

Maybe this error is specific for Windows or for Python 3.8, I have not investigated this. But moving the `webserver` function out from the locals helped.

test/test.py
AndreyMZ: The same as above. I encountered an error on running the test:
~~~
AttributeError: Can't pickle local object 'TestSend.setUp.<locals>.webserver'
~~~

test/test.py
Ekleog: @kaashif Oh, sorry for being mistaken about that!

@AndreyMZ Hmm… so if it doesn't point to the module but to the executable indeed, then the source of the issue is the changes like
```diff
-command = [r2e_path, "-VVVVV",
+command = [sys.executable, r2e_path, "-VVVVV",
```
Where you're calling `python {r2e executable}`, which isn't correct if r2e is being wrapped for any reason. And seeing as `R2E_PATH` was designed to take the wrapped r2e, this can't work.

So… just removing the `R2E_PATH` environment variable (both from `nix/release.nix` and the `os.getenv` in `test/util/execcontext.py`) should be enough to solve the issue, seeing as here we'd be specifically wanting a non-wrapped version of r2e? Either that, or revert the changes to the commands run, so that they don't call `python {r2e exec}` but directly `{r2e exec}`.

What do you think about these solutions?

test/test.py
kaashif: This is interesting, macOS Python 3.8 gets the same error. I've seen at least one other similar issue: https://github.com/pytorch/pytorch/issues/36515. Something to do with multiprocessing changing in 3.8 from using fork to spawn on macOS and I guess this probably never worked on Windows.

test/test.py
AndreyMZ: > revert the changes to the commands run, so that they don't call `python {r2e exec}` but directly `{r2e exec}`.

This would prevent running the tests on Windows.

>  just removing the R2E_PATH environment variable

Given that nobody needs `R2E_PATH` environment variable, I think this is the best solution.

I was hoping that I wouldn't have to deal with the configuration of Travis CI and Nix by myself... OK, I tried, but I have not succeed in running this CI locally. So, I have no idea how to run/debug it after changing `nix/release.nix` before commit. (I prefer not to commit the code which I never run.) Could someone help me with this, please?

Ekleog: Of course! Once nix is installed ([instructions here](https://nixos.org/download.html)), you can just run `nix-build nix/release.nix -A pythonVersions` :)

AndreyMZ: Done. I have removed the `R2E_PATH` environment variable, see https://github.com/rss2email/rss2email/pull/112/commits/c43715d2c86c8105d2b410ffb91a6cf3a3a7b400.

Ekleog:
Great, thank you! (also, thank you for your clean git history, it made everything much easier to review)

There's just 05779b7cd65d07f426d006dccd251199ff666aaa that I'm not sure about (skipping the last .0 doesn't mean we're going to not do semver, and I'm not sure we actually want to have to think of updating to have the .dev suffix), but I've merged for now anyway, hoping to be able to trigger a release with these changes :)

Ekleog:
Ok, release passed, there was just a small hitch related to this PR (the test/util/ folder wasn't getting into the release, which led to needing bf35f09836fa4149baa7a3e3714567acfbc3b272 as well as maybe b19cf8029d5ec1550c33d8bf9fa00fb2d6af0dfb, though I'm not sure about this second one because I thought I had rebased it out and discovered only after having validated all the checks and uploaded to pypi that it had been left in)

Thank you for this great piece of work!

PR #114, opened 2020-05-30
Fix opmlexport output file writing

### LINKED ISSUE #113: opmlexport does not work
The `opmlexport` command try to open the output file into read mode, cf. https://github.com/rss2email/rss2email/blob/master/rss2email/command.py#L178, and of course it cannot write into it after that or it fails because the file does not exist.

nikaro:
Fixes #113

### DIFF
--- a/rss2email/command.py
+++ b/rss2email/command.py
@@ -175,17 +175,17 @@ def opmlexport(feeds, args):
     "Export configuration to OPML."
     if args.file:
         _LOG.info('exporting feeds to {}'.format(args.file))
-        f = open(args.file, 'rb')
+        f = open(args.file, 'wb')
     else:
         _LOG.info('exporting feeds to stdout')
         f = _sys.stdout
     f.write(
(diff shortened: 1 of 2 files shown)


auouymous:
It should probably use "wb" since opmlimport uses "rb".

nikaro:
It is writing a text based file, so AFAIK using binary mode is not useful. It does not prevent the export file to be imported. And if i used `wb` i should have rewrite all `write('...')` to `write(b'...')`, so a bigger diff.

But as you want, let me know if if must change it to `wb`.

Ekleog:
Thank you for this PR! I also think it would be slightly better to use `wb`, as AFAIU there could be newline changes if `w` only and a few added `b` should not be a big deal :)

Would it be possible for you to also add a regression test, so that we can't miss it if `opmlexport` ever gets broken this bad again?

nikaro:
It's done. Let me know if there is something else to do.

Ekleog:
Looks great, thank you! :)

PR #118, opened 2020-06-06
Add GitHub action to build and run tests on Ubuntu, Windows and macOS

kaashif:
This change comes after I saw #112 where a bug completely broke rss2email on Windows. We should run the tests on Windows and macOS to make sure this doesn't happen, since I'm only testing on Debian. I didn't do this through Travis since Travis doesn't support Python on Windows or macOS.

After this is merged, I think we should get rid of Travis CI completely, since it would be redundant and is in fact strictly worse than what is provided by GitHub actions. Also, my opinion is that the CI is much easier to understand if it's literally just `pip install -r requirements.txt; ./test/test.py -v` from a clean machine, since this is what I'm doing locally.

To see what a run of this looks like, see here for an example: https://github.com/kaashif/rss2email/runs/745685834

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest]
         python-version: [3.5, 3.6, 3.7, 3.8]
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,0 +1,31 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


kaashif:
Actually, I realised the tests completely fail on Windows and fail for Python 3.8 on macOS: https://github.com/kaashif/rss2email/runs/745685834. This is despite passing on Ubuntu. @AndreyMZ already fixed one of the bugs in #112 involving pickling working differently on Linux (...I think that's what the problem is, anyway, I can't reproduce it without getting a Windows VM probably).

kaashif:
Despite the failures, I guess this PR is still correct. Fixing the bugs for non-Linux should come later. Not sure if we even have any users on macOS, but we do have at least one on Windows.

Ekleog:
Hmm I'm not sure we want the CI to be red, as it'd mean that until all the bugs are fixed all PRs will show CI failing.

Maybe remove macos and windows from this PR, thus adding just the linux github actions, and then later on we could do PRs that enable windows and fix all the issues?

kaashif:
So I guess the order of operations is:

* Merge this PR but just for Linux
* Now we can see if the CI build works for #112, currently it fails for some nix-related reason
* Merge #112 which has the fixes for Windows
* Re-enable Windows builds and have 6 green ticks!

Sounds good? I'll take a look at #112 soon.

Ekleog:
Sounds good to me!

FWIW, I've just looked into why the nix build was failing, and it looks #112 does have some issues, even if it would maybe pass CI (I don't know) — going to comment there :)

PR #122, opened 2020-06-28
add List-{ID,Post} headers to emails

pacien:
This patch populates some standard List headers as defined in RFC2919 and RFC2369.
Those are commonly recognised by some MUAs, allowing automatic tagging and sorting.

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -478,6 +478,8 @@ class Feed (object):
                 ('Date', self._get_entry_date(entry)),
                 ('Message-ID', '<{}@dev.null.invalid>'.format(_uuid.uuid4())),
                 ('User-Agent', self.user_agent),
+                ('List-ID', '<{}.localhost>'.format(self.name)),
+                ('List-Post', 'NO (posting not allowed on this list)'),
                 ('X-RSS-Feed', self.url),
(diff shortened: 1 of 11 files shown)


pacien:
Travis CI seems to fail due to some lockfile issue which shouldn't be related to the changes proposed here.
https://travis-ci.com/github/rss2email/rss2email/builds/173386418#L4245

rss2email/feed.py
```diff
@@ -478,6 +478,9 @@ def _process_entry(self, parsed, entry):
                 ('Date', self._get_entry_date(entry)),
                 ('Message-ID', '<{}@dev.null.invalid>'.format(_uuid.uuid4())),
                 ('User-Agent', self.user_agent),
+                ('List-ID', '<{}.localhost>'.format(self.name)),
+                ('List-Archive', '<{}>'.format(self.url)),
```
Ekleog, line 482: I'm just not sure about `List-Archive`, as here it doesn't really point to the archives, but just to the feed. I'm also not sure what benefit adding this header brings?

pacien: It's not really useful and it's not pointing to a proper archive indeed.
I removed it.

Ekleog:
Thank you! :)

PR #128, opened 2020-08-23
Set log level from configuration

ryneeverett:
The verbose setting specified in the configuration file doesn't seem to
actually set the log level anywhere.


Ekleog:
LGTM, could you just add a test case to make sure that we don't regress this? (also, if you could drop a line in the changelog in addition to that, it'd be great)

Ekleog (approved):
LGTM, thank you! Unfortunately it looks like this PR conflicts with the changes introduced in #129 in the changelog file, so I can't just merge it now :/

ryneeverett:
Rebased.

Ekleog:
Thank you :)

PR #129, opened 2020-08-23
Add sendmail_config option

ryneeverett:
This takes a path to a sendmail configuration file in case one does not
want to use the global configuration.


Ekleog:
This makes sense to me. Can you just add a changelog entry for this change? :)

Ekleog:
LGTM, thank you! (feel free to send a message after pushing to a branch, github doesn't send notifications upon a (force-)push, so I hadn't noticed you had handled my initial review :))

PR #130, opened 2020-08-23
Log sendmail output

ryneeverett:
There's really no way to debug a broken sendmail setup when the output
is thrown away. Stdout and stderr were passed to the exception as
attributes but do not appear to be used there.

I don't believe there's any other good way to merge stdout/stderr and
preserve the sequential order than to pass one to the other and the
other to PIPE.

### DIFF
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -38,6 +38,7 @@ from email.header import Header as _Header
 from email.mime.text import MIMEText as _MIMEText
 from email.utils import formataddr as _formataddr
 from email.utils import parseaddr as _parseaddr
+import logging
 import mailbox as _mailbox
 from email.utils import getaddresses as _getaddresses
 import imaplib as _imaplib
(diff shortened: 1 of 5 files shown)


Ekleog:
Could you add a unit test, to give an example of how your PR improves things? (feel free to override sendmail with whatever that breaks and burn)

Also, it'd probably be best to add a changelog entry, like “give better logs when sendmail fails” or similar

ryneeverett:
I rebased on top of #128 and #129 because I needed the sendmail_config option and the captured stderr for testing.

ryneeverett:
Rebased.

rss2email/email.py
```diff
@@ -360,12 +360,12 @@ def sendmail_send(recipient, message, config=None, section='DEFAULT'):
         p = _subprocess.Popen(
             sendmail + ['-F', sender_name, '-f', sender_addr, recipient],
             stdin=_subprocess.PIPE, stdout=_subprocess.PIPE,
-            stderr=_subprocess.PIPE)
-        stdout,stderr = p.communicate(message_bytes)
+            stderr=_subprocess.STDOUT)
+        stdout, _ = p.communicate(message_bytes)
         status = p.wait()
+        _LOG.error(stdout.decode())
```
Ekleog, line 366: This should happen only when sendmail actually fails. This being said, it'd probably make sense to have `_LOG.info(stdout.decode())` for cases where sendmail succeeds.

Does that make sense to you?

Ekleog (changes requested):

rss2email/email.py
ryneeverett: When sendmail is successful there is zero output so there's no reason to suppress sendmail output regardless of the log level. An example:

```sh
$ echo "From: me@myhost.com                                                                                                                                                              
To: user@example.com
Hi, this is my message, and I'm sending it to you!
." | sendmail user@example.com
$ echo $?                                                                                                                                                                                
0
```

Ekleog: Well, we'd still display one error line I think:
```python
>>> import logging
>>> logging.error("")
ERROR:root:
```
(also, I wouldn't be surprised if some other sendmail implementations did output things even in successful cases, as sendmail is a “binary” that actually has at least one implementation per SMTP server)

ryneeverett: Force pushed an update that should resolve this concern.

Ekleog (approved):
LGTM :) going to merge if no one objects to it in a bit

Ekleog:
Merging. Thank you! :)

PR #131, opened 2020-08-23
Make read_configfiles a local variable

ryneeverett:
It doesn't seem to be used anywhere else in the code base.


Ekleog:
LGTM, thank you :)

PR #135, opened 2020-08-31
Force the From address by default

gfa:
Blog authors may get bounces with the default configuration, this changes the default configuration to send bounces to the email address configured (user@rss2email.invalid by default)

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=968875 for the original report.

### DIFF
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -85,7 +85,7 @@ CONFIG['DEFAULT'] = _collections.OrderedDict((
         ('use-8bit', str(False)),
         # True: Only use the 'from' address.
         # False: Use the email address specified by the feed, when possible.
-        ('force-from', str(False)),
+        ('force-from', str(True)),
         # True: Use the publisher's email if you can't find the author's.
         # False: Just use the 'from' email instead.
         ('use-publisher-email', str(False)),
--- a/test/data/bbc-chinese/2.config
+++ b/test/data/bbc-chinese/2.config
@@ -2,3 +2,4 @@
 to = a@b.com
 date-header = True
 use-8bit = True
+force-from = False
--- a/test/data/bbc-chinese/3.config
+++ b/test/data/bbc-chinese/3.config
@@ -1,4 +1,4 @@
 [DEFAULT]
 to = a@b.com
 date-header = True
-force-from = False
+use-8bit = True
--- a/test/data/bbc-chinese/3.expected
+++ b/test/data/bbc-chinese/3.expected
@@ -1,0 +1,46 @@
+SENT BY: =?utf-8?b?YmJjY2hpbmVzZS5jb20gfCDkuLvpobU6IEJCQyBDaGluZXNl?= <user@rss2email.invalid>
+MIME-Version: 1.0
+Content-Type: text/plain; charset="utf-8"
+From: =?utf-8?b?YmJjY2hpbmVzZS5jb20gfCDkuLvpobU6IEJCQyBDaGluZXNl?= <user@rss2email.invalid>
+To: a@b.com
+Subject: =?utf-8?b?5Zu96ZmF6LSn5biB5Z+66YeR57uE57uH56ew5YWo55CD57uP5rWO5oGi5aSN5pS+57yT?=
+Content-Transfer-Encoding: base64
+Date: Wed, 23 Jan 2013 20:50:57 -0000
+Message-ID: <...@dev.null.invalid>
+User-Agent: rss2email/...
+List-ID: <test.localhost>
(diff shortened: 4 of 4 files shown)


Ekleog:
Cool, thank you! Could you just change the `r2e.1` file to show that `force-from` defaults to `True`, and maybe add a changelog line so people know we've changed a default setting? (seeing as the default setting being what it is was was probably a bug I don't think it deserves bumping to 4.0, even though technically we're breaking backwards-compatibility… but the change certainly deserves a changelog line nevertheless :))

auouymous:
How would this break compatibility? Wouldn't all upgrades continue to use false for the setting and remain "broken" forever?

The proper fix is for sendmail_send() to always use the `from` config option when calling sendmail. The same would need to be done for other delivery methods.

The `force-from` config option could then be removed since the `use-publisher-email` option does the same thing.

```
diff --git a/rss2email/email.py b/rss2email/email.py
index a4028d1..c8b8efc 100644
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -351,11 +351,12 @@ def sendmail_send(sender, recipient, message, config=None, section='DEFAULT'):
     message_bytes = _flatten(message)
     sendmail = config.get(section, 'sendmail')
     sender_name,sender_addr = _parseaddr(sender)
+    envelope_name,envelope_addr = _parseaddr(config.get(section, 'from'))
     _LOG.debug(
         'sending message to {} via {}'.format(recipient, sendmail))
     try:
         p = _subprocess.Popen(
-            [sendmail, '-F', sender_name, '-f', sender_addr, recipient],
+            [sendmail, '-F', envelope_name, '-f', envelope_addr, recipient],
             stdin=_subprocess.PIPE, stdout=_subprocess.PIPE,
             stderr=_subprocess.PIPE)
         stdout,stderr = p.communicate(message_bytes)
```

Ekleog:
Changing the default value means in particular that when the line is not in the file (as ISTR it currently is not in a default-generated file) the behavior will switch to using true for the setting.

People who manually set force-from to true would still get this behavior, but presumably it is what they wanted — removing the force-from setting, IMO, is something for which we would have to wait for rss2email v4.0, as we would be completely removing any way of recovering the old behavior.

auouymous:
Okay, I undersand it now, after more testing. The feed item author email is always used if available and `use-publisher-email = True` falls back to channel webMaster when author isn't found. The `force-from` setting never uses the author or webMaster email, overriding any value in `use-publisher-email`. The comment in config.py for `use-publisher-email` doesn't make this clear for the true case and has it entirely wrong for the false case, suggestly that `from` is used.

Using `force-from` to fix this issue does break the current behavior of showing the feed email in the `From:` field. And anyone who wants this behavior would turn off `force-from` and completely undo this broken solution.

So `force-from` can't be removed, but it shouldn't be the fix to this issue. Always using `from` for the envelope sender should be the solution and the default value for `force-from` would remain unchanged. There wouldn't be any backward compatible issues. Feeds without an email already use the `from` setting for the envelope. And there is never a valid case where the feed email should be used for the envelope.

And my config file contains every setting. I can remove any setting, including `force-from`, and running r2e puts the removed settings back in the config file. Old config files would have `force-from = False` and continue using the broken behavior.

Ekleog:
Hmmm as I just never use r2e to change the config file, I had completely forgotten that running r2e puts everything back in there. That's unfortunate, as it means that this fix isn't actually enough in practice :(

But as you say, anyway the proper fix is to always set the envelope/5321 from to be the `from` setting, with the same default of `user@rss2email.invalid` as currently probably makes most sense — this way, the inner/5322 from can stay the same as currently, and it reduces the backwards-compat breakage (the people who have filters on `Received` headers would be warned by the changelog, but that feels marginal enough to just warrant a changelog entry.

Thank you for the explanation! I hadn't understood initially that this was your suggested solution :)

This being said, I don't think your patch is correct, because it doesn't fix SMTP sending. Instead, I think the fix ought to be in `_process_entry` in `rss2email/feed.py`, keeping its returned `message` the same but having its returned `sender` be the fixed version — hopefully this'd fix all sending methods, by still having `message` be the RFC5322 message but `sender` be the RFC5321 sender.

auouymous:
I mentioned that my patch didn't handle the other delivery methods, but it is correct, excluding the line I forgot to remove.

The minimal patch would be to pass `self.from_email` to _email.send() in feed.py. Modifying sender earlier than that would break _send_digest(), as it uses sender for the `From:` field in message. Removing the sender parameter from smtp_send(), sendmail_send() and send() guarantees a future mistake won't revert this change.

Issue #134, opened 2020-08-31
 rss2email forges envelope sender


gfa:
From Debian's BTS

```
Today I learned that rss2email copies the email addresses from feed
entries into both the From field and the envelope sender of messages.

This is not acceptable behaviour in an email generator.  The envelope
sender *must* be sent to an address that the user configures, where
*they* can receive bounce messages.

The current behaviour results in bounces being sent to the authors of
feed entries, which is what just happened to me.  It can also result
in messages being dropped if the forgery is detected by MTAs that
check SPF.
```
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=968875

this is an example I found among the feeds I'm subscribed to

```
Received: (nullmailer pid 337350 invoked by uid 3008);
        Mon, 31 Aug 2020 08:24:09 -0000
MIME-Version: 1.0
Content-Type: text/html; charset="utf-8"
Content-Transfer-Encoding: base64
From: "Musings on Life in Hong Kong: StephenC" <noreply@blogger.com>
To: gfa@zumbi.com.ar
Subject: Remote Service-Learning (rSL, eSL)
Date: Mon, 31 Aug 2020 08:24:09 -0000
Message-ID: <2897958a-128d-43c7-aa45-4de1b297fed0@git>
User-Agent: rss2email/3.12.1 (https://github.com/rss2email/rss2email)
List-ID: <hker1.localhost>
List-Post: NO (posting not allowed on this list)
X-RSS-Feed: https://stephencfchan.blogspot.com/feeds/posts/default
X-RSS-ID: tag:blogger.com,1999:blog-11154691.post-8455113540037839365
X-RSS-URL: https://stephencfchan.blogspot.com/2020/08/remote-service-learning-rsl-esl.html
X-RSS-TAGS: Cambodia,education,eSL,Rwanda,service-learning,Vietnam
```

I found a similar bug, with a patch, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=500860

thanks!

Ekleog:
Thank you for this report! Is this problem not already solved by the `force-from` configuration option?

If `force-from` indeed already solves the issue, maybe the best way forward would be to just make it be `True` by default?

PR #137, opened 2020-09-04
Always use the 'from' setting for smtp and sendmail envelope.

auouymous:
Email addresses from the RSS feed must never be used as the envelope `From` address. Doing so causes bounces to be sent to feed authors instead of the user running rss2email.

Fixes https://github.com/rss2email/rss2email/issues/134

Replaces https://github.com/rss2email/rss2email/pull/135

### DIFF
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -151,7 +151,7 @@ def get_message(sender, recipient, subject, body, content_type,
             message[key] = _Header(value, encoding)
     return message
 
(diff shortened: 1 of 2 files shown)


Ekleog:
LGTM, thank you!

Ekleog:
@auouymous Oh, I forgot asking you for a changelog line that states that we no longer forge envelope senders. Could you send in a PR with such a change?

PR #138, opened 2020-09-04
Remove unused TimeoutError message parameter.

auouymous:
TimeoutError is tightly coupled to TimeLimitedFunction, making it
unlikely the message parameter would ever be used by it or anything
else.


Ekleog:
Thank you!

Issue #136, opened 2020-09-03
rss2email issue regarding XDG_RUNTIME_DIR and lockfile


Ekleog:
(Here is a copy of an email, from an email from `Lucas <lucas@sexy.is>` who doesn't have a github account. I haven't reviewed it in-depth, and it includes a patch that should solve the issue raised)

Hey,

As we talked on IRC, attached you can find a reproduction script for
the issue.

To recap:
- rss2email/main.py:172 creates a lockfile
- Base directory for the lockfile is XDG_RUNTIME_DIR, otherwise /tmp
- lockfile name is the same for every user
- As long as users don't set XDG_RUNTIME_DIR, lockfile is always
  /tmp/rss2email.lock, making it impossible to use `r2e` after another
  user used it

Possible solutions / workarounds:
1. Use a personalized, per user directory when XDG_RUNTIME_DIR isn't
   set: for example, tmux uses `/tmp/tmux-$uid` with permissions `0700`.
2. Tell the user to always use XDG_RUNTIME_DIR

For 1., check the attached patch. After extracting it, you can apply it
running, inside the repository:

    git am 0001-Create-lockfile-in-an-user-specific-directory.patch

That will keep my authorship in place. It's a PR in mail, basically :P
Patch is made against commit 34c924173a8858e3dd8aa914d53f77e3e43fefc8,
current master as of time of writing. If you need to create a PR
anyways from it, you can apply the patch and push against a branch, and
create the PR from that branch.

Cheers!
-Lucas

`[ 0001-Create-lockfile-in-an-user-specific-directory.patch: text/plain ]`
```diff
From 80c29040fecc40871f460b972f59b1508e218374 Mon Sep 17 00:00:00 2001
Message-Id: <80c29040fecc40871f460b972f59b1508e218374.1599017054.git.lucas@sexy.is>
From: Lucas <lucas@sexy.is>
Date: Wed, 2 Sep 2020 03:15:25 +0000
Subject: [PATCH] Create lockfile in an user-specific directory

Current approach makes `r2e` create the lockfile under
`/tmp/rss2email.lock` if XDG_RUNTIME_DIR is unset, for all users,
making it impossible to use in shared shell environments. Instead,
create the lockfile under `/tmp/rss2email-${UID}/rss2email.lock`.

While there, document that XDG_RUNTIME_DIR is used for storing the
lockfile.
---
 r2e.1             | 4 ++++
 rss2email/main.py | 7 ++++++-
[...]

PR #142, opened 2020-09-11
Create lockfile in an user-specific directory

Ekleog:
Current approach makes `r2e` create the lockfile under
`/tmp/rss2email.lock` if XDG_RUNTIME_DIR is unset, for all users,
making it impossible to use in shared shell environments. Instead,
create the lockfile under `/tmp/rss2email-${UID}/rss2email.lock`.

While there, document that XDG_RUNTIME_DIR is used for storing the
lockfile.

### DIFF
--- a/r2e.1
+++ b/r2e.1
@@ -302,6 +302,10 @@ $HOME/\&.config.
 The preferred directory for data files.  Defaults to
 $HOME/\&.local/share.
 .TP
+.B XDG_RUNTIME_DIR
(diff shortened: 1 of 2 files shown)


Ekleog:
This is the PR from #136

kaashif:
Looks good to me (and also to @auouymous according to https://github.com/rss2email/rss2email/issues/136#issuecomment-691061504), so I'll merge this.

PR #148, opened 2020-09-19
In feed titles, replace newlines with spaces, to avoid header parsing errors

smokris:
When I try to send emails for a feed that has a linebreak in its title, I get an exception.

For example, https://sarahcandersen.com/rss starts with:

```xml
<rss xmlns:dc="http://purl.org/dc/elements/1.1/" version="2.0"><channel><description></description><title>Sarah's
Scribbles</title><generator>Tumblr (3.0; @sarahseeandersen)</generator>…
```

There's a linebreak between the word "Sarah's" and the word "Scribbles", which results in this exception:

```
  File "/usr/bin/r2e", line 5, in <module>
    rss2email.main.run()
  File "/usr/lib/python3.6/site-packages/rss2email/main.py", line 171, in run
  File "/usr/lib/python3.6/site-packages/rss2email/command.py", line 87, in run
    feed.run(send=args.send)
  File "/usr/lib/python3.6/site-packages/rss2email/feed.py", line 849, in run
    for (guid, id_, sender, message) in self._process(parsed):
  File "/usr/lib/python3.6/site-packages/rss2email/feed.py", line 367, in _process
    processed = self._process_entry(parsed=parsed, entry=entry)
  File "/usr/lib/python3.6/site-packages/rss2email/feed.py", line 507, in _process_entry
    section=self.section)
  File "/usr/lib/python3.6/site-packages/rss2email/email.py", line 132, in get_message
    sender_name = str(_Header(sender_name, sender_encoding).encode())
  File "/usr/lib64/python3.6/email/header.py", line 390, in encode
    "an embedded header: {!r}".format(value))
email.errors.HeaderParseError: header value appears to contain an embedded header: "Sarah's\nScribbles: <author>"
```

This PR replaces the linebreak with a space, enabling the feed to be successfully emailed.


auouymous:
Works for me.

amiryal (changes requested):
I’d include the sample feed in `test/data/`. It would help to protect against future regressions, and also make it easier for other contributors to hack on different solutions.

You can also expand the commit message, to explain the issue in more detail.

rss2email/feed.py
```diff
@@ -654,6 +654,7 @@ def _get_entry_name(self, parsed, entry):
         if 'name' in feed.get('publisher_detail', []):
             data['publisher'] = feed.publisher_detail.name
         name = self.name_format.format(**data)
+        name = name.replace('\n', ' ')
```
amiryal, line 657: To make it consistent with this other line:
https://github.com/rss2email/rss2email/blob/9f6ca0af6de81845dd68a1330d45d2345ee422e6/rss2email/feed.py#L594
Add `.strip()` at the end. They both seem to serve the same purpose.

amiryal (changes requested):

Ekleog:
Thank you for this PR!

Just, as @amiryal mentioned, could you maybe add a test here? Also, if it were possible to also add a changelog entry, I think this would be good to go :)

Ekleog:
@smokris Are you still interested in pushing this to completion?

smokris:
Thanks for the feedback, and my apologies for not getting back to this sooner.  I just pushed a new version, which includes the `.strip()` change, and adds a test and changelog entry.

amiryal (approved):

Ekleog:
Awesome thank you!

PR #149, opened 2020-09-21
Support feedparser 6.

### LINKED ISSUE #73: Migrate to newer feedparser
It seems to me like feedparser has the same issue we did: they have over 100 unreleased commits since their last release in 2015 including fixes for breaking issues like this one. Too bad it's considered impolite to harass people into regularly releasing their code.

_Originally posted by @kaashif in https://github.com/rss2email/rss2email/issues/72#issuecomment-528318277_

auouymous:
Closes #73

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,8 @@ UNRELEASED
     * Stop forging SMTP and sendmail envelope sender (#134)
     * Add sendmail_config option
     * Fix verbose configuration setting
+    * Upgraded to feedparser v6.0 (https://github.com/kurtmckee/feedparser/)
+    * Drop support for Python 3.5
 
 v3.12.2 (2020-08-31)
     * Fix bug `AttributeError: 'NoneType' object has no attribute 'close'` (#126)
--- a/HACKING.md
+++ b/HACKING.md
@@ -43,9 +43,9 @@ against multiple python versions.
 You can build each one of them like this:
 
 ```
-nix-build -A pythonVersions.rss2email-python_3_5 nix/release.nix
 nix-build -A pythonVersions.rss2email-python_3_6 nix/release.nix
 nix-build -A pythonVersions.rss2email-python_3_7 nix/release.nix
+nix-build -A pythonVersions.rss2email-python_3_8 nix/release.nix
 …
 ```
 
--- a/nix/release.nix
+++ b/nix/release.nix
@@ -3,7 +3,6 @@ let
   pkgs = import nixpkgs {};
 
   supportedPackageSets = [
-    { version = "3_5"; set = pkgs.python35Packages; }
     { version = "3_6"; set = pkgs.python36Packages; }
     { version = "3_7"; set = pkgs.python37Packages; }
     { version = "3_8"; set = pkgs.python38Packages; }
@@ -31,7 +30,7 @@ let
     '';
   };
 
-  # { "rss2email-python_3_5" = <rss2email package>; … }
+  # { "rss2email-python_3_6" = <rss2email package>; … }
   rss2emailVersions =
     (pkgs.lib.listToAttrs
       (map
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-feedparser==5.2.1
+feedparser>=6.0.0
 html2text==2020.1.16
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -325,7 +325,7 @@ def _flatten(message):
     """
     bytesio = _io.BytesIO()
     # TODO: use policies argument instead of policy set in `message`
-    # see https://docs.python.org/3.5/library/email.generator.html?highlight=bytesgenerator#email.generator.BytesGenerator
+    # see https://docs.python.org/3.6/library/email.generator.html?highlight=bytesgenerator#email.generator.BytesGenerator
     generator = _BytesGenerator(bytesio)
     try:
         generator.flatten(message)
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -433,7 +433,7 @@ class Feed (object):
         elif isinstance(exc, _SOCKET_ERRORS):
             _LOG.error('{}: {}'.format(exc, self))
             warned = True
-        elif isinstance(exc, _feedparser.zlib.error):
+        elif isinstance(exc, _feedparser.http.zlib.error):
             _LOG.error('broken compression: {}'.format(self))
             warned = True
         elif isinstance(exc, (IOError, AttributeError)):
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,6 @@ setuptools.setup(
         'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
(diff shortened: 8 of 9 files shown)


Ekleog:
Hmm… I think that for this to pass CI, requirements.txt / setup.py must be updated too?

auouymous:
@Ekleog Are you able to see the fail logs, or are they unavailable for you as well?

Ekleog:
I see this, right now:
```
======================================================================
ERROR: test_email_data/disqus/1.config (__main__.TestEmails)
----------------------------------------------------------------------
  File "./test/test.py", line 61, in fn
    self.run_single_test(config_path=test_path)
  File "./test/test.py", line 150, in run_single_test
    '\n'.join(diff_lines)))
ValueError: error processing data/disqus/1.config
--- expected
+++ generated
@@ -43,7 +43,7 @@
 List-ID: <test.localhost>
 List-Post: NO (posting not allowed on this list)
 X-RSS-Feed: data/disqus/feed.rss
-X-RSS-ID: ab03f2100069a1cd0876b997be87976c18d48e8a
+X-RSS-ID: a52375ec78a988241fe9864a2243d4d910538d52
 X-RSS-URL: http://software-carpentry.org/2012/11/who-wants-to-write-a-little-code/#comment-713578640
```

(on the 3.7 one, the two others are cancelled but eg. 3.6 looks pretty similar to me)

If you can't read them… I guess there's some permission wrong in github maybe? Or maybe it was just a temporary issue on the github side?

auouymous:
I get that same error locally with feedparser 5.2.1 and unpatched rss2email, and also with feedparser 6.0.1 and patched rss2email. I was told to ignore it, and for some reason it never caused github tests to fail until now.

The unavailable logs was my excessive UBO blocking, fixed now.

Ekleog:
Hmm… how are you testing? When I test locally, on master, by doing `nix-shell --run 'python test/test.py'`, I don't get any error. However, when testing with this additional patch on top of your PR (to use feedparser 6.0.1, which didn't make it upstream in nixpkgs yet):
```
diff --git a/shell.nix b/shell.nix
index d891a2d..621085d 100644
--- a/shell.nix
+++ b/shell.nix
@@ -20,7 +20,25 @@ let
 in pkgs.mkShell {
   buildInputs = with pythonPackages; [
     pkgs.python3
-    feedparser
+    (feedparser.overridePythonAttrs (old: rec {
+      pname = "feedparser";
+      version = "6.0.1";
+      src = fetchPypi {
+        inherit pname version;
+        sha256 = "0k77c66ia4m64kmdfa2yv8z8887hhwx91pv8b4s2ix23mbf8xa3c";
+      };
+      propagatedNativeBuildInputs = [
+        (buildPythonPackage rec {
+          pname = "sgmllib3k";
+          version = "1.0.0";
+          src = fetchPypi {
+            inherit pname version;
+            sha256 = "1s8jm3dgqabgf8x96931scji679qkhvczlv3qld4qxpsicfgns3q";
+          };
+          doCheck = false;
+        })
+      ];
+    }))
     html2text
     update-copyright pkgs.git
   ];
```
I get this:
```
......E........<?xml version="1.0" encoding="UTF-8"?>
<opml version="1.0">
<head>
<title>rss2email OPML export</title>
</head>
<body>
</body>
</opml>
<?xml version="1.0" encoding="UTF-8"?>
<opml version="1.0">
<head>
<title>rss2email OPML export</title>
</head>
<body>
<outline type="rss" text="test" xmlUrl="https://example.com/feed.xml"/>
</body>
</opml>
...
======================================================================
ERROR: test_email_data/disqus/1.config (__main__.TestEmails)
----------------------------------------------------------------------
  File "test/test.py", line 61, in fn
    self.run_single_test(config_path=test_path)
  File "test/test.py", line 150, in run_single_test
    '\n'.join(diff_lines)))
ValueError: error processing data/disqus/1.config
--- expected
+++ generated
@@ -43,7 +43,7 @@
 List-ID: <test.localhost>
 List-Post: NO (posting not allowed on this list)
 X-RSS-Feed: data/disqus/feed.rss
-X-RSS-ID: ab03f2100069a1cd0876b997be87976c18d48e8a
+X-RSS-ID: a52375ec78a988241fe9864a2243d4d910538d52
 X-RSS-URL: http://software-carpentry.org/2012/11/who-wants-to-write-a-little-code/#comment-713578640

 @Hans-Martin

----------------------------------------------------------------------
Ran 18 tests in 3.924s

FAILED (errors=1)
```

So I guess the CI, despite not using nix, is having the same results as I do locally with nix?

FWIW, I've also pushed to master a patch bumping its nixpkgs version, so that I'm not testing with ancient python dependencies when doing the tests with nix, but I can confirm the above results still reproduce with nixpkgs-20.09 — everything working fine on master, but one failing test on a branch with your PR.

My guess would be that feedparser 6 just changed the way it computes IDs in some circumstances, in which case I'm not sure whether it means we must release rss2email 4 or not… but maybe we should anyway, seeing as feedparser is a major portion of rss2email's logic and had extensive years of work that went into it?

auouymous:
I test with `env "PATH=$(pwd):$PATH" "PYTHONPATH=$(pwd):$PYTHONPATH" ./test/test.py -v`, which uses whatever feedparser Gentoo has installed. That failed X-RSS-ID is the same with feedparser 5.2.1, 6.0.0 and 6.0.1. It is the same as what I had in April https://github.com/rss2email/rss2email/pull/103#issuecomment-610157306.

> That test failure is a long-standing intermittently flaky test, due to the orders of attributes in the HTML changing between runs (see #11), changing the hash. Don't worry about that particular test failure. We might have to fix this soon though, contributors always ask about this.

Ekleog:
Hmm… I guess before the issue stemmed from some other python package in your python path, then? What I can say is:
- the test failed all three runs (3.6, 3.7 and 3.8) here
- the test did not fail any of the three runs on other recent PR's CI runs (eg. https://github.com/rss2email/rss2email/pull/140 for the most recent one)
- the test repeatably succeeds with nix before your PR and repeatably fails with nix after your PR (tested on 10 runs each).

So, to me it doesn't look like just flakiness, but like your previous local test failure was due to some difference in your global environment, that led to this happening on your local computer but neither on mine (because nix) nor on CI (because it starts with an empty image).

And, now, with the feedparser upgrade, somehow this difference you had in your global environment now made it into the development environment (maybe because it's in some dependency of feedparser, that now sees its version bumped?).

Does this analysis look sound to you?

If so, I'd say the best fix would be to edit the test to now have this newer X-RSS-ID, this way we can land this without breaking CI for all subsequent PR's. Does that make sense to you?

auouymous:
It appears feedparser uses sgmllib which was removed in python 3. Gentoo patches sgmllib code into feedparser 5 and depends on external sgmllib3k for 6. Without sgmllib, all ops become pass, and all non-Gentoo users with python 3 were/are running feedparser 5 without sgmllib.

The following change to rss2email/feed.py causes the test to work.

```
     def _get_entry_hash(self, entry) -> str:
         content = self._get_entry_content(entry)
-        content_value = content['value'].strip()
+        content_value = content['value'].strip().replace('<br />', '<br>')
```

The problem is how feedparser (with sgmllib) terminates void elements, changing `<br>` and `<br/>` to `<br />`. The test feed contains `<br>` and the allthingsrss feed contains `<br/>`, but its tests work because they have real GUIDs and don't rely on hashing the content.

Ekleog (approved):
Great, thank you! Can you just add a line to the changelog, so it's clear that one of our dependencies underwent a major update? And then it's good to land, I think :)

auouymous:
Done.

In https://github.com/rss2email/rss2email/pull/149/commits/ad183796aed0ed600326252a5948b9c2fc234507, is the `>=` correct for requirements.txt or should it be `==`. Also, html2text has two different versions visible in that commit, should the version in setup.py be changed in another PR?

Ekleog:
Hmm… I think `~= 6.0` would be best here instead of `>=` and `==`, does that sound good to you too?

As for html2text, you're right, we certainly ought to at least make `setup.py` and `requirements.txt` be in sync, and probably this time use `==` as 2020.1.16 doesn't look very semver-y.

This being said… maybe just using `>=` everywhere would actually be best? this way CI will fail and people will complain when we stop having the latest versions, which actually may be a good thing as I'd guess otherwise we'll eventually end up with ancient versions without ever noticing it…

Either way, IMO this is good enough to merge, it just has that merge conflict issue on the changelog — I guess this PR is based off a version of master that predates the landing of a few PRs with changelog changes a few days ago?

auouymous:
Rebased.

Ekleog:
Great, thank you!

PR #151, opened 2020-10-10
Poetry2nix

ryneeverett:
Currently the tests are broken in nix-shell because feedparser was
updated to depend on 6.0.0 which is not yet in nixpkgs. This resolves
that issue and will make dependency resolution easier to manage going
forward.


Ekleog:
LGTM, thank you!

PR #153, opened 2020-10-12
Add option to remove seen entries no longer in the feed.

### LINKED ISSUE #49: Add db cleanup feature to prevent heavy databases
Hi,

I reported in the former project that I remarked that json db tends to grow significantly. (Original issue: https://github.com/wking/rss2email/issues/33). I experienced that it slows down quite heavily the run command.

I proposed a workaround here: https://github.com/wking/rss2email/pull/56
It basically reset and run to get only the lastest seen feeds in the db. I use this command regularly without any issue although, there is a little risk that I miss an entry if the feed gets updated between the run and the cleandb commands. A more careful work must be done on this, but I never found the time to get it done.

I report here again as I believe this issue is major in my use case.

Thank you for reading and thank you for keeping this beautiful software alive!

auouymous:
Using `r2e run --clean` will remove seen entries no longer found in the feed, avoiding the race-condition with `r2e reset ; r2e run -n`. This option could also be used instead of `r2e run` to keep the database small, or used periodically due to it being a little slower.

Some feeds only show the N most recent items and if a recent item is removed from the feed it will cause an older item to reappear, and r2e will resend it. The second patch keeps the 3 most recent old entries and should be enough to never resend an old item.

Should there be a short -C option, or another letter since it resembles -c?

Fixes #49

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@ UNRELEASED
       $XDG_DATA_HOME
     * Warnings about HTTP content-type being unexpected now properly display
     * Make the proxy parameter also affect https connections
+    * Add a --clean argument on the run command to reduce the database size
 
 v3.12.2 (2020-08-31)
     * Fix bug `AttributeError: 'NoneType' object has no attribute 'close'` (#126)
--- a/r2e.1
+++ b/r2e.1
@@ -67,7 +67,7 @@ optional \fI<email>\fR argument is the email address to send new items
 to, overriding the default address for this particular feed.  Repeat
 for each feed you want to subscribe to.
 .TP
-.B run \fR[\fI\-\-no-send\fR] \fR[\fI<index>\fR [\fI<index>\fR ...]]
+.B run \fR[\fI\-\-no-send\fR] \fR[\fI\-\-clean\fR] \fR[\fI<index>\fR [\fI<index>\fR ...]]
 Scan the feeds and send emails for new items. This can be run in a cron
 job.
 .P
@@ -75,6 +75,10 @@ job.
 The \-\-no-send option stops \fBr2e\fR from sending any email. This can be
 useful the first time you run it, as otherwise it would send an email
 for every available feed entry.
+.P
+The \-\-clean option reduces the database size by removing old entries. It
+forces a download of selected feeds or all feeds, and should only be used
+weekly or monthly.
 .P
 If an \fI<index>\fR is specified, \fBr2e\fR will only download that
 feed. \fI<index>\fR can be either the feed name (as set by \fBadd\fR)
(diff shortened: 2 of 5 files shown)


auouymous:
Now it shouldn't crash when not cleaning, and won't reset feeds without changes that support etag.

Ekleog:
This idea looks good to me! About the implementation however, for the “keep most recent 3 entries,” I wonder… it looks to me like you're taking the 3 removed entries with lowest guid, instead of with the highest timestamp, am I misreading something?

auouymous:
There is no per-entry timestamp in the `seen` list.

A feed with 3 entries is added to `seen` as [1,2,3] where 3 is the most recent entry. On the next check, a fourth entry is appended to the end of `seen` as [1,2,3,4]. There is no shuffling or sorting of the `seen` list, its order is always the order entries were released. The patch walks backwards through the list, keeping the first 3 old entries and discarding old entries nearest the list head.

Ekleog:
Oh. I just understood why this works: `dict` officially keeps the insertion order since python 3.7, which makes `list` return the order in which things were added to `seen`. Could you maybe just add a comment around the `list(self.seen)` so it's easier to understand while reading the code?

So, the things I think are missing from this PR:
* Dropping official support for python 3.6 in this PR, as AFAIU 3.6 did not specify that `list(d)` returned the keys in the insertion order into `d`. IMO such a change can land anyway, because we're switching to feedparser 6 anyway so the next release will be a big one.
* Adding a changelog entry
* Adding documentation in the manpage

Does that sound reasonable to you?

Ekleog:
Oh and also: it would be most awesome if you could also add a test that will help avoid regressions, but OTOH I don't think that's required to land it, as I don't think we have the infrastructure yet to make such a test easy to implement.

auouymous:
After you read the comments in latest patch, should I still bump the python version? It is possible non-CPython implementations below 3.7 don't order the list, but most of the feature will still work. Non-CPython 3.6 users, who use --clean, would only get a resent entry if the feed removed an entry and the next oldest wasn't one of the three kept. Python 3.6 isn't EOL until December and Feedparser 6 still supports it, I don't think we should drop support.

A test would need to add a feed from file, run clean using a second feed file that has fewer entries, and then verify rss2email.json has the correct entries.

auouymous:
Also, a non-CPython 3.6 implementation could have saved the unordered entries and upgrading to 3.7 could still resend if the feed removes a new entry. That could happen a year or more from now for low activity feeds. So forcing them to upgrade now won't fix the issue, it only reduces the chance for high activity feeds.

Ekleog:
Looks great to me! You're right, I hadn't thought of the fact that the order of insertion was taken into account at first viewing time and not at cleanup time, so let's say that we're still 3.6-compatible and land this, thank you :)

Issue #101, opened 2020-03-06
Support for “multipart/alternative”


mawis:
Currently rss2email either sends the e-mail messages in plain text or in HTML, depending on the presence of the setting `html-mail = True`.

I don't read the messages generated by rss2email always on the same device. Actually when I read them on my mobile I much prefer the HTML formatting – while reading the mails on my desktop PC (using `mutt`) I much prefer the text encoding.

Normally with e-mails this is no problem as the typical standard is to send e-mails in `multipart/alternative` encoding having one part in plain text and the other in HTML. With rss2email this currently doesn't seem to be possible, while rss2email is already able to generate both encodings.

I'd like to see either a setting, that both encodings using `multipart/alternative` are used for sending e-mails or to automatically include a plain text alternative if rss2email is set up to send out HTML e-mails.

Ekleog:
Done with #158, thank you @boyska!

PR #158, opened 2020-12-24
#101: make HTML emails actually multipart/alternative

boyska:
This pull request adds support for creating `multipart/alternative` emails, thus fixing #101.

This support can be enabled with the `multipart-html` boolean option (defaults to False, see below).

It works like this:
 * it does NOT modify `text/plain` messages
 * whenever a `text/html` message is generated, a new `multipart/alternative` message is created. It preserves all the headers, add a mimepart for `text/html` from the original message and generates a new `text/plain` part using `html2text`
 * given the way rss2email works, digest messages are properly transofmed: each message in the digest is now multipart/alternative.

I set `multipart-html` default to False for two reasons:
 1. not to break current tests
 2. not to change current behavior unexpectedly

Apart for these two reasons, I think that multipart/alternative is always preferable over raw text/html.

I don't know what are your policies about integration tests. I'd like to add some tests to make sure that `multipart-html` option is robust, but I don't know what's your workflow to create expected values for tests. Shall I just run it, look if I like the result, and add it to `test/data/` directory? If so, I can do it!

## Thanks

Adding support for multipart emails that are both nicer and widely supported will contribute to a [better outreach of Tails project](https://gitlab.tails.boum.org/tails/tails/-/issues/16723)! That's because Tails, a portable operating system that protects against surveillance and censorship, uses rss2email for its newsletter. Thanks for it!

### DIFF
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -36,6 +36,7 @@ from email.generator import BytesGenerator as _BytesGenerator
 import email.header as _email_header
 from email.header import Header as _Header
 from email.mime.text import MIMEText as _MIMEText
+from email.mime.multipart import MIMEMultipart
 from email.utils import formataddr as _formataddr
 from email.utils import parseaddr as _parseaddr
 import logging
@@ -49,6 +50,8 @@ import subprocess as _subprocess
 import sys as _sys
 import time as _time
 import os as _os
+
+import html2text
 
 from . import LOG as _LOG
 from . import config as _config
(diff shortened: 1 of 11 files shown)


rss2email/email.py
```diff
@@ -76,6 +79,34 @@ def guess_encoding(string, encodings=('US-ASCII', 'UTF-8')):
...
+    text_part = _MIMEText(text_content)
+    alternative.attach(text_part)
+    return msg
+
+def message_add_plain_multipart(guid, message, html):
+    if message.get_content_type() == 'text/html':
+        m = _add_plain_multipart(guid, message, html)
+        return m
+    if message.is_multipart():
+        # we could support multipart messages, but let's postpone it
```
Ekleog, line 106: Maybe log a warning message?

Ekleog (commented):
Looks good to me! I think this PR is just missing tests and a changelog entry. For the tests, the procedure you described is perfect: run it, look if you like the result, and if so add it to the `test/data/` directory — what's important is that the result is always good, not necessarily that it's always exactly something defined beforehand.

Thank you for this PR!

rss2email/email.py
boyska: sounds like the right thing to do. In fact, I don't expect any message arriving in that part of the code to be a multipart message.

boyska:
> Looks good to me! I think this PR is just missing tests and a changelog entry.

Done! I added a test case to the "gmane" suite. I also added an entirely new feed, which is the Tails one. I was looking for some feed that had html formatting, links, etc. :)
If you think there's no need for it, removing it is trivial so just ask!

boyska:
Hello! Would you mind giving a look at it again? Let me know if I can do anything to improve it, of course!

Ekleog:
This looks awesome, thank you! Sorry about the delay for review, I kinda unexpectedly became maintainer of rss2email a while ago, and so it's not exactly the first thing I'm thinking of when doing code for fun, which leads to github notifications sometimes taking month to be answered… sorry about that!

boyska:
> This looks awesome, thank you! Sorry about the delay for review, I kinda unexpectedly became maintainer of rss2email a while ago, and so it's not exactly the first thing I'm thinking of when doing code for fun, which leads to github notifications sometimes taking month to be answered… sorry about that!

no worry for the delay, and thanks for maintaining rss2email!

PR #159, opened 2020-12-28
Simplify test case creation

boyska:
While working on #158 I found it hard to generate the `*.expected` files. So I looked at `test.py` and modified it:
 - `run_single_test` has no `force`argument anymore. I don't know if there was any case in which it was passed, indeed.
 - the equivalent of it is the environment variable `FORCE_TESTDATA_CREATION` ; set it to `1` and you are now in "force mode"
 - in force mode, only _missing_ files are created. existing files are not overwritten

The workflow to create a test case is now much simpler:
 - create `1.config` insde `data/mysuite/`
 - run `env FORCE_TESTDATA_CREATION=1 python3 -m unittest`. This will show you which files have been created, reporting them as errors.
 - Now you can run `python3 -m unittest` and it should pass
 - Of course, you need to check if you like the result
 - You might also need to anonymize Message-IDs

Hope you like it!

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -142,7 +142,7 @@ class TestEmails(unittest.TestCase, metaclass=TestEmailsMeta):
                     f.write(generated)
                 raise ValueError('missing expected test data, now created')
             else:
-                raise ValueError('missing test; set FORCE_TESTEDATA_CREATION=1 to create')
+                raise ValueError('missing test; set FORCE_TESTDATA_CREATION=1 to create')
         else:
             with open(expected_path, 'r') as f:
                 expected = self.clean_result(f.read())
--- a/test/test.py
+++ b/test/test.py
@@ -116,13 +116,13 @@ class TestEmails(unittest.TestCase, metaclass=TestEmailsMeta):
             text = regexp.sub(replacement, text)
         return text
 
-    def run_single_test(self, dirname=None, config_path=None, force=False):
+    def run_single_test(self, dirname=None, config_path=None):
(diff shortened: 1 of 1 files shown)


amiryal (approved):
I was also working on a patch and didn’t understand how to generate a new `.expected` file without modifying `if force:` to `if True:` temporarily. In the case of creating a new `.expected`, your proposed changes make it straightforward. It becomes trickier when you need to update an existing file, in which case you would have to delete the old one before running with `FORCE_TESTEDATA_CREATION=1`. This is good enough, though, so LGTM.

test/test.py
```diff
@@ -131,15 +131,21 @@ def run_single_test(self, dirname=None, config_path=None, force=False):
...
-                f.write(generated)
+
+        expected_path = config_path.replace('config', 'expected')
+        if not _os.path.exists(expected_path):
+            if _os.environ.get('FORCE_TESTDATA_CREATION', '') == '1':
+                with open(expected_path, 'w') as f:
+                    f.write(generated)
+                raise ValueError('missing expected test data, now created')
+            else:
+                raise ValueError('missing test; set FORCE_TESTEDATA_CREATION=1 to create')
```
Ekleog, line 145:
```suggestion
                raise ValueError('missing test; set FORCE_TESTDATA_CREATION=1 to create')
```

Ekleog (approved):
This looks great to me too, thank you! I'll add a few lines to the readme after landing, so it's easier to know about this :)

PR #166, opened 2021-03-18
Fix double escaping of sender names on certain configurations

amiryal:
Before this patch, if the sender name included non-ASCII characters, and 'US-ASCII' was not the first in the user-configurable list of encodings, then the sender name would be doubly escaped, and the MUA would dutifully display an unfriendly From: =?utf-8?b?…?= <…@…> (see full explanation in the commit message). The fix is to let `formataddr` do its job and not to re-process its result.

The included new test case demonstrates nicely how the badly-encoded message looks like, when run with the old code.


Ekleog:
Looks great to me, thank you! :)

PR #172, opened 2021-03-19
Add wrap-links option to prevent links from wrapping over multiple lines

### LINKED ISSUE #96: URLs line-wrapped and therefore broken when using text format emails

auouymous:
Closes #96


Ekleog:
Awesome, thank you!

PR #174, opened 2021-03-26
Perform HTML escaping on text to be appended to HTML

amiryal:
The first commit, which I am trying to sneak in by the back door, is a follow up on this remark by @boyska:
> * You might also need to anonymize Message-IDs
>  — https://github.com/rss2email/rss2email/pull/159#issue-546105927

It does the cleanup of Message-IDs and everything else at the time of creating the test case, and saves the extra work of doing the cleanup on every file read afterwards.

The second commit adds a test case where one of the entry titles contains an ampersand `&` character. When the plain text subject is concatenated as is into the HTML message, the results is not valid HTML. In the patch, I tried to find all places where entity escaping is necessary, although only the subject is covered by the test case.

NB: I found the quiltville feed by a random search for `site:blogspot.com` on Duckduckgo and looking for an ampersand in the title. If anyone would like to suggest a different feed, I wouldn’t mind.


rss2email/feed.py
```diff
@@ -803,15 +803,15 @@ def _process_entry_content(self, entry, content, subject):
             if self.use_css and self.css:
                 lines.extend([
                         '    <style type="text/css">',
-                        self.css,
+                        _saxutils.escape(self.css),
```
Ekleog, line 806: Whoever reading this and doubting it's correct because self.css is a user-configurable and not feed-provided: I think this is correct because one can escape CSS inline styles and it still works… or so I think.

Ekleog (approved):
LGTM, thank you!

PR #175, opened 2021-03-26
Set HTML attribute dir=auto on body element

amiryal:
Has no effect on non-RTL text, or on feed items that have their `dir` attribute already set properly, but can help greatly with reading RTL items that neglect to specify the `dir` attribute.

See [the spec](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute) for some hard-core technicalities, or scroll down to the example there for a visual demonstration.

I am opening this as a draft PR to gather comments, but will wait with rebasing until #174 is decided because of conflicting changes.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@ UNRELEASED
     * Warnings about HTTP content-type being unexpected now properly display
     * Make the proxy parameter also affect https connections
     * Add a --clean argument on the run command to reduce the database size
+    * Set body element attribute dir=auto in HTML mail
 
 v3.12.2 (2020-08-31)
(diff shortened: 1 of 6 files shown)


Ekleog:
Ugh so I'm learning that it's not the default… yet another item of europeano-centrism in our tech stack :(

This LGTM, thank you! As #174 landed, can you rebase, unmark as draft, and just add a line to the changelog so people are aware of it, as IIUC it has the potential of sometimes breaking stuff?

amiryal:
> IIUC it has the potential of sometimes breaking stuff?

Given a predominantly RTL paragraph, it has the potential of not doing the right thing, but that is a great improvement over nearly always doing the wrong thing. ☺

Ekleog:
Awesome, thank you! :D

PR #177, opened 2021-04-03
Updated required Python version in README

jpalus:


amiryal (approved):
The `README.rst` change LGTM. As for `poetry.lock`, I don’t know the Right Way to effect changes in it, but it seems safe enough to edit by hand, to change `python-versions = ">=3.5"` to `python-versions = ">=3.6"` to match, no?

jpalus:
Not sure if the question is to me or not :). Anyway I'm not using poetry myself but it looks like `python-versions = ">=3.5"` in poetry.lock refers to html2text dependency no?

amiryal:
> it looks like `python-versions = ">=3.5"` in poetry.lock refers to html2text dependency no?

You are right, I was being too lazy with my `git grep`. Sorry for the noise!

Ekleog:
LGTM, thank you! For the poetry.lock, I've just pushed a commit updating it, though AFAICT it shouldn't need a backport to release :)

PR #181, opened 2021-04-10
Fix feedparser status check regression on Python>=3.9

### LINKED ISSUE #178: [3.13|master] 5 tests fail with Python 3.9 (but not 3.7 or 3.8)
5 tests seem to fail for Python 3.9 (but not 3.7. or 3.8) with both release 3.13 and Git `HEAD` on `master`:

```console
# cd "$(mktemp -d)"
# git clone --depth 1 https://github.com/rss2email/rss2email
# cd rss2email/
# git rev-parse HEAD
9a508aa0e0af37c749fcc4d3d446f49aecd2add9
# virtualenv --python=python3.9 venv
# source venv/bin/activate
# pip install setuptools .
# cd test/
# python --version
Python 3.9.2
# python -m unittest --verbose test_reply_changes.py
test_default (test_reply_changes.TestReplyChanges) ... FAIL
test_no_send (test_reply_changes.TestReplyChanges) ... FAIL
test_off (test_reply_changes.TestReplyChanges) ... FAIL
test_on (test_reply_changes.TestReplyChanges) ... FAIL
test_switch (test_reply_changes.TestReplyChanges) ... FAIL

======================================================================
FAIL: test_default (test_reply_changes.TestReplyChanges)
----------------------------------------------------------------------
  File "/tmp/tmp.mPUH6QldWF/rss2email/test/test_reply_changes.py", line 34, in test_default
    self.assertEqual(1, len(messages))
AssertionError: 1 != 0

======================================================================
FAIL: test_no_send (test_reply_changes.TestReplyChanges)
----------------------------------------------------------------------
  File "/tmp/tmp.mPUH6QldWF/rss2email/test/test_reply_changes.py", line 92, in test_no_send
    self.assertEqual(2, len(messages))
AssertionError: 2 != 0

======================================================================
FAIL: test_off (test_reply_changes.TestReplyChanges)
----------------------------------------------------------------------
  File "/tmp/tmp.mPUH6QldWF/rss2email/test/test_reply_changes.py", line 43, in test_off
    self.assertEqual(1, len(messages))
AssertionError: 1 != 0

======================================================================
FAIL: test_on (test_reply_changes.TestReplyChanges)
----------------------------------------------------------------------
  File "/tmp/tmp.mPUH6QldWF/rss2email/test/test_reply_changes.py", line 52, in test_on
    self.assertEqual(2, len(messages))
AssertionError: 2 != 0

======================================================================
FAIL: test_switch (test_reply_changes.TestReplyChanges)
----------------------------------------------------------------------
  File "/tmp/tmp.mPUH6QldWF/rss2email/test/test_reply_changes.py", line 76, in test_switch
    self.assertEqual(2, len(messages))
AssertionError: 2 != 0

----------------------------------------------------------------------
Ran 5 tests in 2.834s

FAILED (failures=5)
```

amiryal:
Fixes #178.

Instead of waiting for kurtmckee/feedparser#272 to be resolved and then updating our requirement from `feedparser>=6.0.0` to the fixed version, I chose to support the already-released versions of feedparser by accepting `None` as a valid status.

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -404,9 +404,12 @@ class Feed (object):
 
     def _check_for_errors(self, parsed):
         warned = False
-        status = getattr(parsed, 'status', 200)
+        status = getattr(parsed, 'status', None)
         _LOG.debug('HTTP status {}'.format(status))
-        if status == 301:
+        if status is None:
+            # On non-HTTP schemes, like file://
+            pass
+        elif status == 301:
             _LOG.info('redirect {} from {} to {}'.format(
                     self.name, self.url, parsed['url']))
             self.url = parsed['url']
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -28,4 +28,5 @@ jobs:
         poetry install
     - name: Test
       run: |
-        poetry run ./test/test.py -v
+        cd ./test
+        poetry run python -m unittest
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}


hartwork (commented):
@amiryal nice work! I'm unsure why `None` is in the set of good values rather than considered an error. If it's the right choice, maybe an in-code comment about why that's the case would be helpful, I think I may not be the last one to find this non-obvious. What do you think?

amiryal:
> I'm unsure why `None` is in the set of good values rather than considered an error. If it's the right choice, maybe an in-code comment about why that's the case would be helpful, I think I may not be the last one to find this non-obvious. What do you think?

Good point @hartwork, an in-code comment was in order! In my latest force-pushed version I didn’t want to burden the code with _all_ of the details, but I hope it’s OK to refer the interested reader to the commit message instead.

rss2email/feed.py
```diff
@@ -404,9 +404,12 @@ def _process(self, parsed):
 
     def _check_for_errors(self, parsed):
         warned = False
-        status = getattr(parsed, 'status', 200)
+        status = getattr(parsed, 'status', None)
         _LOG.debug('HTTP status {}'.format(status))
-        if status == 301:
+        if status is None:
+            # On non-HTTP schemes, like file://. See commit message for further details.
```
hartwork, line 410: Would you mind replacing `See commit message for further details.` with a short version of the actual description? The current approach is rather unusual and some Git users may not even be expert enough about Git to quickly get to that text, when needed. Thank you!

amiryal: I appreciate your thought, but after looking at it again I decided that it’s clear enough without the full explanation. A couple of lines above, `status = getattr(parsed, 'status', None)` obviously might produce a `None`. If anyone cares enough to notice _that_ in their combination of feedparser and Python versions `None` is not possible, then the test matrix will fail them if they try to remove this branch of the `if` statement. They will also see the commit message, if they know where to look, so I removed that sentence eventually.

hartwork (approved):

Ekleog:
As per https://github.com/rss2email/rss2email/pull/188 let's close this as it was fixed upstream, but thank you for your work on this! :)

PR #182, opened 2021-04-26
Add fetch option to add command

timendum:
Hi,
i've added a new option to the `add` command:  `--fetch` (short `-f`), to fetch/initialize a feed while adding.

The as-is alternative is a `add` and a `run --no-send` risking losing any other updates.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,6 +22,8 @@ v3.13 (2021-04-03)
     * Add a --clean argument on the run command to reduce the database size
     * Set body element attribute dir=auto in HTML mail
     * Store the lock file in XDG_RUNTIME_DIR instead of /tmp
+    * New argument `--only-new` on the `add` command to ignore entries in feed
+      when added, so only new entries will be sent.
 
 v3.12.3 (2021-03-19)
     * Make dependency on feedparser have an upper bound so that `pip install`
--- a/r2e.1
+++ b/r2e.1
@@ -60,12 +60,17 @@ sets the default email address that mails are sent to.
 .B email \fR[\fI<email>\fR]
 Update the default target email address to \fI<email>\fR.
 .TP
-.B add \fI<name>\fR \fI<url>\fR [\fI<email>\fR]
+.B add \fR[\fI\-\-only-new\fR] \fI<name>\fR \fI<url>\fR [\fI<email>\fR]
 Subscribe to a feed. The \fI<name>\fR argument gives the feed a name
 for future manipulation. \fI<url>\fR is the URL of the feed.  The
 optional \fI<email>\fR argument is the email address to send new items
 to, overriding the default address for this particular feed.  Repeat
 for each feed you want to subscribe to.
+.P
+.RS 4
+The \-\-only-new option will prevent entries already in the feed to be
+sent.
+.RE
 .TP
 .B run \fR[\fI\-\-no-send\fR] \fR[\fI\-\-clean\fR] \fR[\fI<index>\fR [\fI<index>\fR ...]]
 Scan the feeds and send emails for new items. This can be run in a cron
--- a/r2e.1
+++ b/r2e.1
@@ -68,8 +68,11 @@ to, overriding the default address for this particular feed.  Repeat
 for each feed you want to subscribe to.
 .P
 .RS 4
-The \-\-only-new option will prevent entries already in the feed to be
-sent.
+The \-\-only-new option fetches the feed at addition time without
+sending the entries, making subsequent \fBrun\fR execution consider
+them as already having been sent. It should be used when one is only
+interested in entries that are not yet in the feed at the time of
+running \fBadd\fR.
 .RE
 .TP
 .B run \fR[\fI\-\-no-send\fR] \fR[\fI\-\-clean\fR] \fR[\fI<index>\fR [\fI<index>\fR ...]]
--- a/rss2email/command.py
+++ b/rss2email/command.py
@@ -61,6 +61,8 @@ def add(feeds, args):
     _LOG.info('add new feed {}'.format(feed))
     if not feed.to:
         raise _error.NoToEmailAddress(feed=feed, feeds=feeds)
+    if args.only_new:
+        feed.run(send=False, clean=False)
     feeds.save()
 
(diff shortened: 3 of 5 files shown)


auouymous:
So `r2e add -f <name> <url>` is identical to `r2e add <name> <url> ; r2e run <name> --no-send`? This saves some typing but I don't understand why the later risks losing updates. In either case, you could simply check the feed again to see if anything was released while typing the command.

rss2email/command.py
```diff
@@ -62,6 +62,9 @@ def add(feeds, args):
     if not feed.to:
         raise _error.NoToEmailAddress(feed=feed, feeds=feeds)
     feeds.save()
+    if args.fetch:
+        feed.run(send=False, clean=False)
+        feeds.save()
```
amiryal, line 68: There is only a need to call `feeds.save()` once, at the end:
```python
    if not feed.to:
        raise _error.NoToEmailAddress(feed=feed, feeds=feeds)
    if args.fetch:
        feed.run(send=False, clean=False)
    feeds.save()

```
If we save the configuration file only once _after_ fetching the feed, then this also gives an answer to [@auouymous](https://github.com/rss2email/rss2email/pull/182#issuecomment-827204209): without this patch, there is a race after `r2e add`, between the subsequent interactive `r2e run -n` and the background scheduled `r2e run`. So, this patch would save me both some typing _and_ the need to be careful and look at the clock while adding a feed.

amiryal (changes requested):

timendum:
Thanks @amiryal, I've modified the code and force pushed the change.

amiryal (approved):

auouymous:
Thanks @amiryal, it makes perfect sense now. LGTM.

Ekleog:
Thank you for this PR! While the core idea looks good to me, there is a thing I'm uneasy about: `-f` sounds, to me, like something that has too high a probability of being mistyped. IMO we shouldn't make it easy to confuse the use case “send me an email for each entry already in the feed” and “send me an email only for new entries in the feed”.

While the current default of “send me an email for each entry already in the feed” is debatable, it's the current default and changing it would be a major change that probably should be done as a subsequent change.

So I think that the short name should probably be removed, and the long name replaced with `--only-new` or something similar (better naming ideas welcome)

With this said, this is missing a manpage update, as well as a changelog line. Ideally there'd be a new test that checks the feature continues working after later changes, do you think you'd be able to add one? The test should probably go into `test/test.py` until we split it out, but an example of how to do tests that require changing the feeds can be seen in eg. `test/test_reply_changes.py`.

timendum:
I've changed the option to `--only-new` and added it in the man page.

I'm working on the tests.

timendum:
@Ekleog The pull is now complete, I added one test.

auouymous:
LGTM, other than the following lines.

```
         help="entries in the feed now will be not sent")
```
Transpose 'be not'?

```
     """Span a webserver for `timeout` seconds"""
```
Spawn?

rss2email/main.py
```diff
@@ -98,6 +98,10 @@ def run(*args, **kwargs):
     add_parser.add_argument(
         'email', nargs='?',
         help='target email for the new feed')
+    add_parser.add_argument(
+        '--only-new', dest='only_new',
+        default=False, action='store_const', const=True,
+        help="entries in the feed now will be not sent")
```
Ekleog, line 104:
```suggestion
        help="entries in the feed now will not be sent")
```

test/test.py
```diff
@@ -192,6 +192,19 @@ def webserver_for_test_fetch(queue, num_requests, wait_time):
     finally:
         httpd.server_close()
 
+def webserver_for_test_if_fetch(queue, timeout):
+    """Span a webserver for `timeout` seconds"""
```
Ekleog, line 196:
```suggestion
    """Spawn a webserver for `timeout` seconds"""
```

r2e.1
```diff
@@ -60,12 +60,17 @@ sets the default email address that mails are sent to.
...
+.B add \fR[\fI\-\-only-new\fR] \fI<name>\fR \fI<url>\fR [\fI<email>\fR]
 Subscribe to a feed. The \fI<name>\fR argument gives the feed a name
 for future manipulation. \fI<url>\fR is the URL of the feed.  The
 optional \fI<email>\fR argument is the email address to send new items
 to, overriding the default address for this particular feed.  Repeat
 for each feed you want to subscribe to.
+.P
+.RS 4
+The \-\-only-new option will prevent entries already in the feed to be
+sent.
```
Ekleog, line 72:
```suggestion
The \-\-only-new option fetches the feed at addition time without
sending the entries, making subsequent \fBrun\fR execution consider
them as already having been sent. It should be used when one is only
interested in entries that are not yet in the feed at the time of
running \fBadd\fR.
```

rss2email/main.py
```diff
@@ -98,6 +98,10 @@ def run(*args, **kwargs):
     add_parser.add_argument(
         'email', nargs='?',
         help='target email for the new feed')
+    add_parser.add_argument(
+        '--only-new', dest='only_new',
+        default=False, action='store_const', const=True,
```
Ekleog, line 103:
```suggestion
        '--only-new', action='store_true',
```

Ekleog (approved):
Awesome, thank you! Just committing a few nits on your branch, waiting for CI to complete and this should be ready to go :)

Ekleog:
Well this is good to go. Thank you for your contribution! :D

CHANGELOG
```diff
@@ -22,6 +22,8 @@ v3.13 (2021-04-03)
     * Add a --clean argument on the run command to reduce the database size
     * Set body element attribute dir=auto in HTML mail
     * Store the lock file in XDG_RUNTIME_DIR instead of /tmp
+    * New argument `--only-new` on the `add` command to ignore entries in feed
+      when added, so only new entries will be sent.
```
amiryal, line 26: Oops, added in the wrong place — it should have been added to the UNRELEASED section. @Ekleog watch out for this when you cut the next release.

auouymous: Can't we just push a commit to move it?

PR #186, opened 2021-05-20
Don't rewrite config file when fetching feeds

### LINKED ISSUE #110: rss2email should probably not modify the configuration file upon `r2e run`
Modifying the configuration file upon `r2e run` means that the user allowed to run `r2e` must also be allowed to modify the configuration file.

In quite a few cases, the configuration file would legitimately be owned by `root`, and `r2e` would run under the `rss2email` user, ideally not being allowed to edit the configuration file.

duxovni:
I believe this fixes #110 , it's working fine on my system at least.

### DIFF
--- a/rss2email/command.py
+++ b/rss2email/command.py
@@ -44,7 +44,7 @@ def new(feeds, args):
         feeds.config['DEFAULT']['to'] = args.email
     if _os.path.exists(feeds.configfiles[-1]):
         raise _error.ConfigAlreadyExistsError(feeds=feeds)
-    feeds.save()
+    feeds.save_config()
 
 def email(feeds, args):
     "Update the default target email address"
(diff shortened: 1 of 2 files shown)


auouymous:
LGTM and the config file is not being written to on each run.

amiryal (approved):

Ekleog:
The idea of the change looks great to me but I'm not so sure about the implementation.

Would it be possible to, instead of adding a boolean parameter to `save()`, instead rename `_save_feeds` to `save_feeds` (thus “making it public”), and to change `save` to only save the configuration and rename it `save_config`? And then have each `.save()` call instead call explicitly either `save_config`, `save_feeds`, or both.

This way, the API stays cleaner and more maintainable.

Does that sound reasonable to you? :)

duxovni:
That's a great idea, just pushed that change :)

Ekleog:
Awesome, thank you! :D

Issue #192, opened 2021-07-14
id selectors in css are really config comments


squeakbat:
The default css config value contains id selectors "#entry" and "#body" which are also config file comments.  Not only do they get ignored, they are dropped when the config file is written back.

So, the easy fix is to change them to something like "*#entry"` and "*#body".  However, I think they should be class selectors instead (with appropriate changes in the code), like .header and .footer.  I know this would be an incompatible change, but how many users actually write their own css?  (OK, I do.)

Being classes will also make it possible to implement a feature I want to add: digest as a single-part html message body.  Otherwise, they'll be duplicate ids.

squeakbat:
I should have said I can submit a pull request, whatever is decided here.

Ekleog:
Would it make sense to have them be both class and id attributes in the HTML, and to have the default css use class selectors?

This way we'd avoid the incompatible change, and I think this'd solve the main issue :)

squeakbat:
Pull request #193 merged as commit 4a088c7efe1c76fa1ebe842adc8e2a796df8f313.

PR #193, opened 2021-07-16
Change "entry" and "body" from html id to class.

squeakbat:
Unlike header and footer elements, entry and body were used as ids,
not class.  Not only was this inconsistent, it also didn't actually
work, because the "#entry" and "#body" lines in the default config
file were parsed as comments.

For backward compatibility with users who have defined their own
style sheets, we are keeping them as ids as well
(e.g., &lt;div class="body" id="body">).

This is discussed in issue #192.


Ekleog:
Awesome, thank you! Unfortunately it looks like the quiltville tests didn't pass, FWIW it should be possible to use https://github.com/rss2email/rss2email/#tips-and-tricks to auto-generate the tests :)

squeakbat:
Ugh.  I can't run the tests by hand probably because my python is a little old (whatever comes with debian buster).  So I fixed up the files by hand, clearly not well enough.

I'll take a look, and maybe grab the latest relevant modules so I can test the tests.

Should I redo this whole thing as a new fork and pull request (after it's all straightened out) so it's all in one neat commit.  (I also kind of fucked it up by doing a merge from master.  I hate git.)

auouymous:
You can uncommit, modify and recommit everything locally and then push with `--force` to overwrite all commits in this PR.

Ekleog:
Awesome, thank you! As for improving the git history, it could have been a good thing but GitHub has a “squash and merge” button anyway so there's no big deal here. Thank you for having suggested it anyway!

FWIW, here is the way I'd have done it with raw git, if github didn't have this squash and merge button:
```
$ # Starting from the tip of the PR branch
$ # First, set the current git commit to the parent of your PR without touching the worktree
$ git reset --soft 1fa854eba8dc8342f39a37762b60cfeb95525c29 # the commit on top of which you built your branch, here origin/master would probably have worked too
$ # Now, the worktree hasn't changed and all the previously-made changes are already staged (ie. in the index)
$ # So we just need to actually commit the things as a single commit
$ git commit
[write again the commit message]
$ git push
```

(It might have been possible to do this with `git rebase` too, but given that there is a merge commit in the middle of the PR I'm less sure of that)

Anyway, thank you for the PR! :D

squeakbat:
By the way, thank you all for the help and timely feedback!

PR #198, opened 2021-10-25
Add subject format option

joshaw:
My mail provider prevents setting the `From` address, so I've been finding it difficult to quickly identify which feed a new email is from. This change adds a `subject-format` option that lets you configure the subject line to make that more apparent. I use `subject-format = {feed-name}: {feed-title}` which makes it much easier.

First contribution to the project, so happy to make changes if desired. Thanks.


Ekleog:
This looks great to me, and the feature totally makes sense. Thank you!

PR #199, opened 2021-11-03
Fix manpage formating

jspricke:
"\'" renders as an acute accent and just a "'" at the beginning of a
line is interpreted as Groff command. So reformat the line instead.

### DIFF
--- a/r2e.1
+++ b/r2e.1
@@ -175,8 +175,8 @@ False: Just use the 'from' email instead.
 .IP name-format
 If empty, only use the feed email address rather than
 friendly name plus email address.  Available attributes may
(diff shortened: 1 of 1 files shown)


amiryal (changes requested):
The same commit also added a `\'feed'` at the beginning of [line 231](https://github.com/jspricke/rss2email/commit/63e554c151a70c3e95f646d53784aded12ac8f2c#diff-1d40a883a5d9bc3adec6a3e1084b72c22c0f4bacb871ed18b7a0e4ce23b937ffR231), so need to fix there as well.

jspricke:
done, thx.

Ekleog:
Great, thanks!

PR #202, opened 2022-01-25
Update IRC Channel from freenode to Libera

naclander:
Looks like the new channel is #rss2email at libera.chat


auouymous:
Oops, the channel was moved almost a year ago. Thank you for fixing that.

PR #203, opened 2022-02-04
In case those error occur, the datafile does not exist (.datafile is None), print its path instead.

JulienPalard:
I did not reproduced those (as the require argument is never True, should it be dropped?), I just noticed the issue while reading the code.


auouymous:
I found this same problem while looking into a fix for #204, I'm glad I checked this before posting a duplicate PR. Using `datafile` would only display a worthless `<codecs.StreamReaderWriter object at ...>` if the datafile were not None. And `datafile_path` is used when an external message is passed in.

Thank you for the fix.

amiryal (approved):

PR #205, opened 2022-02-06
Close rss2email.json file for all commands.

### LINKED ISSUE #204: ResourceWarning: unclosed file
Using the environment variables:

    PYTHONDEVMODE=1
    PYTHONTRACEMALLOC=10

I get:

```bash
$ r2e list
0: [*] discourse (https://blog.discourse.org/feed.xml -> julien@palard.fr)
bin/r2e:8: ResourceWarning: unclosed file <_io.BufferedReader name='/home/mdk/.local/share/rss2email.json'>
  sys.exit(run())
Object allocated at (most recent call last):
  File "bin/r2e", lineno 8
    sys.exit(run())
  File "rss2email/main.py", lineno 199
    feeds.load()
  File "rss2email/feeds.py", lineno 233
    self._load_feeds(require=require)
  File "rss2email/feeds.py", lineno 248
    self.datafile = _codecs.open(
  File "/usr/lib/python3.9/codecs.py", lineno 905
    file = builtins.open(filename, mode, buffering)
```

auouymous:
Fixes #204.


rss2email/main.py
```diff
@@ -200,6 +200,7 @@ def run(*args, **kwargs):
         if not args.verbose:
             _LOG.setLevel(feeds.config['DEFAULT']['verbose'].upper())
         args.func(feeds=feeds, args=args)
+        feeds.close()
```
amiryal, line 203: Looking at [the wider context](https://github.com/auouymous/rss2email/blob/e6ff9db705b233f63692610cdbe490ab293d0721/rss2email/main.py#L194-L211):
```python
    try:
        if not args.config:
            args.config = None
        feeds = _feeds.Feeds(datafile_path=args.data, configfiles=args.config)
        if args.func != _command.new:
            feeds.load()
        if not args.verbose:
            _LOG.setLevel(feeds.config['DEFAULT']['verbose'].upper())
        feeds.close()
    except _error.RSS2EmailError as e:
        e.log()
        if _logging.ERROR - 10 * args.verbose < _logging.DEBUG:
            raise  # don't mask the traceback
        _sys.exit(1)
    finally:
        if lockfile is not None:
            lockfile.close()
```
We know now that `if args.func != _command.new:` is redundant (because `_command.new` is not the only exception), and `feeds.close()` could arguably move to the `finally:` clause.

amiryal (changes requested):

rss2email/main.py
auouymous: I agree that `feeds.close()` should be moved to finally. But are you suggesting the `_command.new` conditional is redundant because pause and unpause also don't require feeds to be loaded?

amiryal: I meant it was redundant because we are now closing the file anyway; of course it is better to not open the file in the first place if it is not needed. My [original suggestion](https://github.com/rss2email/rss2email/issues/204#issuecomment-1030767777) was to not call `feeds.load()` here at all, and instead call it only from the places that need it in `rss2email/command.py`. This would be better and more in line with the Zen of Python (“explicit is better than implicit”) than either removing this condition on `args.func != _command.new` or updating it to include all of the exceptions (currently `new`, `email`, `list`, `pause`, `unpause`, `opmlexport`).

auouymous: I agree, it would be better to only load the feeds inside each command that needs them. Do you want to submit a PR for that? Or I can try to get to it in a few days if not.

JulienPalard:
Looks like the lines in the try:

```python
        if not args.config:
            args.config = None
        feeds = _feeds.Feeds(datafile_path=args.data, configfiles=args.config)
```

cannot raise an `RSS2EmailError`, so I would move it out of the `try:`.

Keeping it in the try may lead someone in the future to raise an RSS2EmailError from the __init__ (or a called function from the __init__), leading to a `UnboundLocalError` on `feeds.close()`.

If I'm wrong and those line can raise an `RSS2EmailError`, we already have a possible `UnboundLocalError` on `feeds.close()`.

amiryal:
> cannot raise an `RSS2EmailError`, so I would move it out of the `try:`.

The purpose of the `try…finally` block is to release the lockfile, so I wouldn’t move the `feeds` initialisation out of there. @JulienPalard I added a check `if feeds is not None:`, so I think it can be merged now.

JulienPalard (approved):

amiryal:
I retrospectively thought about not having updated the changelog, but I guess it’s OK if the change does not affect end users.

auouymous:
Sorry about the delay on this, and thank you @amiryal for fixing the patch.

PR #207, opened 2022-02-18
Add docker image

rgibert:


MarcelRobitaille:
I had to change the `Dockerfile` like so:

```Dockerfile
FROM python:3-slim AS build-env
ENV PIP_DISABLE_PIP_VERSION_CHECK=1
COPY . /app
WORKDIR /app
# Pip install not needed. Handled by setup.py install
RUN python setup.py install

FROM gcr.io/distroless/python3
WORKDIR /app
COPY --from=build-env /app /app
# Proper site packages location
# Determined by import site; site.getsitepackages()
# Otherwise, I get "no module feedparser"
COPY --from=build-env /usr/local/lib/python3.10/site-packages /usr/lib/python3/dist-packages
# `ENTRYPOINT [ "/app/r2e", "run" ]` game me `no such file or directory /app/r2e` even
# though the file definitely existed
# `ENTRYPOINT [ "python", "/app/r2e", "run" ]` works though

# I have volumes in my docker-compose.yml for data and config
# Data has to go in a subdirectory. Using the file directly as a volume gives OSError: [Errno 16] Device or resource busy: '/app/database.json.tmp' -> '/app/database.json'
# Probably the file is open because of how it's mounted
ENTRYPOINT [ "python", "/app/r2e", "--data", "/app/data/database.json", "--config", "/app/config.cfg", "run" ]
```

I know you said in #218 that this is not ready. May I ask what is missing? I would like to contribute. I plan to run it like this, so I can report any issues.

MarcelRobitaille:
It's common to run rss2email using cron, it's even covered in the docs, so I think there should be a cron version as well.

`Dockerfile`:
```Dockerfile
FROM python:3.10.7-alpine3.16
COPY . /app
WORKDIR /app
RUN python setup.py install
CMD [ "/app/cron.sh" ]
```

`cron.sh`:
```bash
#!/bin/sh
crontab /app/crontab
crond -f
```

In this case, you also have to pass `crontab` as a volume to `/app/crontab`. This way, it is possible to configure the schedule outside of the container.

Issue #120, opened 2020-06-20
user-agent setting has an unintuitive name: it sounds like HTTP user agent, but it's RFC5322 user agent


nikaro:
The User-Agent set in the configuration is not passed to the `feedparser.parse()` function. Is it intended?

This does the job:
```
diff --git a/rss2email/feed.py b/rss2email/feed.py
index 5641bc1..ff8a6cf 100644
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -365,7 +365,8 @@ class Feed (object):
         if proxy:
             kwargs['handlers'] = [_urllib_request.ProxyHandler({'http':proxy})]
         f = _util.TimeLimitedFunction(timeout, _feedparser.parse)
-        return f(self.url, self.etag, modified=self.modified, **kwargs)
+        return f(self.url, etag=self.etag, modified=self.modified,
+                 agent=self.user_agent, **kwargs)

     def _process(self, parsed):
         _LOG.info('process {}'.format(self))
```

Ekleog:
@nikaro Hmm… can you exhibit a case where the user-agent does not get passed in the HTTP request? or, even better, a test? If so, if you can submit a PR I'll be glad to accept it :)

auouymous:
Config file has `user-agent = rss2email` and the web server log shows this.

```
"GET /test-rss.php HTTP/1.1" 200 711 "-" "UniversalFeedParser/5.2.1 +https://code.google.com/p/feedparser/"
```

And here is the log after applying that rebased patch.

```
"GET /test-rss.php HTTP/1.1" 200 713 "-" "rss2email"
```

Ekleog:
Gotcha, thank you :) so I guess we probably just need some regression test to land this

nikaro:
Sorry unfortunately i don't use rss2email anymore.

Ekleog:
With the discussion in #154 and re-reading #74, I've just noticed this was actually a mistake as per documentation in the manpage: `user-agent` is behaving as expected, even though the name is misleading.

Renaming the issue to track the fact that the issue is actually that this setting has an unintuitive name, and keeping #154 open to track the feature that this issue initially wanted

Ekleog:
Idea for fixing: add a setting `mail-user-agent` that is an alias to current `user-agent`, and add a deprecation warning to `user-agent`. #154 can then add an `http-user-agent` setting

amiryal:
Nitpicking with regards to the title:

[...]

PR #208, opened 2022-03-05
Pass user-agent to feedparser

rand0mdud3:
Hi there, would it be possible to pass the user agent configured to feedparser. Have a bunch of feeds that report 403 without this.


auouymous:
This is a duplicate of #120 which was determined to be incorrect. https://github.com/rss2email/rss2email/issues/120#issuecomment-723130855 has the proper fix for this, but needs someone to implement it.

Issue #154, opened 2020-11-06
option for change User-Agent in http headers


kjonca:
Some sites refuses fetching rss, when user-agent is not known to them. I prepared simple change which allow define "http-user-agent-option" and pass it to parser (diff is against debian version but I think this is no issue)

kjonca:
```

--- rss2email-3.12.2.orig/rss2email/config.py
+++ rss2email-3.12.2/rss2email/config.py
@@ -100,6 +100,7 @@ CONFIG['DEFAULT'] = _collections.Ordered
         ### Fetching
         # Set an HTTP proxy (e.g. 'http://your.proxy.here:8080/')
         ('proxy', ''),
+        ('http-user-agent', ''),
         # Set the timeout (in seconds) for feed server response
         ('feed-timeout', str(60)),
         # Set the time (in seconds) to sleep between fetches from the same server
--- rss2email-3.12.2.orig/rss2email/feed.py
+++ rss2email-3.12.2/rss2email/feed.py
@@ -75,8 +75,9 @@ from . import email as _email
 from . import error as _error
 from . import util as _util
 
-
 _urllib_request.install_opener(_urllib_request.build_opener())
+
+ 
 _SOCKET_ERRORS = []
 for e in ['error', 'herror', 'gaierror']:
     if hasattr(_socket, e):
@@ -370,6 +371,11 @@ class Feed (object):
             config = self.config['DEFAULT']
         proxy = config['proxy']
         timeout = config.getint('feed-timeout')
+
+        http_user_agent = config ['http-user-agent']
+        if http_user_agent:
+            _feedparser.USER_AGENT= http_user_agent
+
         kwargs = {}
         if proxy:
             kwargs['handlers'] = [_urllib_request.ProxyHandler({'http':proxy})]

```

auouymous:
Is this different from the existing `user-agent` setting? The patch in #120 fixes it.

kjonca:
According to documentation "user-agent" is used in outgoing mails, not http headers.  Even if that patch, make it used in http headers, I am  not sure if this value is proper for both purposes (so I used different option name)

auouymous:
#120 was waiting for someone to write a regression test, but you are correct about the documentation saying `user-agent` is for outgoing mails. So #120 is probably a new feature and not a bug fix.

[...]

PR #209, opened 2022-03-05
Add a new http-user-agent config.

rand0mdud3:
Gives the ability to override the HTTP user agent.

This should take care of https://github.com/rss2email/rss2email/issues/154

Also related to https://github.com/rss2email/rss2email/issues/120 (which applies to the mail user agent).

### DIFF
--- a/r2e.1
+++ b/r2e.1
@@ -260,6 +260,9 @@ Path to sendmail (or compatible)
 .IP user-agent
 String to use as User-Agent in outgoing emails. If present, __VERSION__ and
 __URL__ are replaced with rss2email version number and webpage
+.IP http-user-agent
+String to use as HTTP User-Agent in web requests. If present, __VERSION__ and
+__URL__ are replaced with rss2email version number and webpage
 .RE
 .SS SMTP configuration
 .IP smtp-auth
--- a/rss2email/config.py
+++ b/rss2email/config.py
@@ -88,6 +88,8 @@ CONFIG['DEFAULT'] = _collections.OrderedDict((
         ('from', 'user@rss2email.invalid'),
         # The User-Agent default string (rss2email __VERSION__ and __URL__ is replaced)
         ('user-agent', 'rss2email/__VERSION__ (__URL__)'),
+        # The HTTP User-Agent default string (rss2email __VERSION__ and __URL__ is replaced)
+        ('http-user-agent', 'rss2email/__VERSION__ (__URL__)'),
         # Transfer-Encoding. For local mailing it is safe and
         # convenient to use 8bit.
         ('use-8bit', str(False)),
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -187,6 +187,9 @@ class Feed (object):
     _default_configured_attributes[
         _default_configured_attributes.index('user_agent')
         ] = '_user_agent'  # `user_agent` is a getter that does substitution
+    _default_configured_attributes[
+        _default_configured_attributes.index('http_user_agent')
+        ] = '_http_user_agent'  # `http_user_agent` is a getter that does substitution
     # all attributes that are saved/loaded from .config
     _configured_attributes = (
         _non_default_configured_attributes + _default_configured_attributes)
@@ -236,6 +239,12 @@ class Feed (object):
     @property
     def user_agent(self):
         return self._user_agent.\
+            replace('__VERSION__', __version__).\
+            replace('__URL__', __url__)
+
+    @property
+    def http_user_agent(self):
+        return self._http_user_agent.\
             replace('__VERSION__', __version__).\
             replace('__URL__', __url__)
 
@@ -384,7 +393,7 @@ class Feed (object):
                 _urllib_request.ProxyHandler({ 'http': proxy, 'https': proxy })
             ]
         f = _util.TimeLimitedFunction('feed {}'.format(self.name), timeout, _feedparser.parse)
-        return f(self.url, self.etag, modified=self.modified, **kwargs)
+        return f(self.url, self.etag, modified=self.modified, agent=self.http_user_agent, **kwargs)
 
     def _process(self, parsed):
         _LOG.info('process {}'.format(self))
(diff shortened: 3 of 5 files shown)


auouymous:
LGTM. However, it was decided to deprecate `user-agent` and alias it to `mail-user-agent`, but I'm not sure how to do that. @Ekleog Do you know how?

> Idea for fixing: add a setting mail-user-agent that is an alias to current user-agent, and add a deprecation warning to user-agent. #154 can then add an http-user-agent setting

amiryal (changes requested):
Seems like the old `user_agent` config already _was_ used for HTTP, in the optional redirect filter:

https://github.com/rss2email/rss2email/blob/853ec71b6e595b372c8b63c61d2a24aa9cdc29bc/rss2email/post_process/redirect.py#L66

Can you change it to use the new `http_user_agent`?

I found it while drafting a solution for [adding a `mail-user-agent` option](https://github.com/rss2email/rss2email/pull/209#issuecomment-1060112489), BTW. When I have time, and if the branch allows it, I will push my work onto here.

Ekleog:
@auouymous 
> LGTM. However, it was decided to deprecate user-agent and alias it to mail-user-agent, but I'm not sure how to do that. @Ekleog Do you know how?

We're probably doing a release soon as the changes are slowly piling up, so I'd say something like:
1. Add a mail-user-agent option and make previous uses of user-agent use mail-user-agent instead
2. Have the old user-agent option just set mail-user-agent and output a warning that on the next release the user-agent option will be removed and people should already migrate to mail-user-agent
3. Update the documentation
4. Add the change to the changelog
5. Make the release
6. Make user-agent error out instead of warning+redirecting to mail-user-agent

Does that make sense?

rand0mdud3:
Apologies for the lag, fixed that use case I missed, thanks @amiryal!

amiryal:
> I found it while drafting a solution for [adding a `mail-user-agent` option](https://github.com/rss2email/rss2email/pull/209#issuecomment-1060112489), BTW. When I have time, and if the branch allows it, I will push my work onto here.

That line of work :point_up: did not get anywhere useful. Since it was already overloaded for use in HTTP, and since I can’t think of a use case for 2 different values for HTTP and message-header agent strings*, I think we can kidnap `user-agent` and use it for both. (I.e. neither `http-user-agent` nor `mail-user-agent` should be added, and instead use the existing `user-agent` config for HTTP.)

---

\* Frankly, I can’t think of a use case for changing the message header at all. It seems like this one is purely informational for the end recipient, so it’s only useful to change it if one likes to hide the use of r2e for whatever reason. If that is the case, then we can direct users to write a custom post_process hook.

amiryal:
> neither `http-user-agent` nor `mail-user-agent` should be added, and instead use the existing `user-agent` config for HTTP

@rand0mdud3 if you agree with the above, then I think it would be OK to implement it in this PR. Just remember to reflect the additional use of the config option in the manual.

auouymous:
I agree with not using user-agent for HTTP and instead adding a new http-user-agent, but I don't fully agree with removing user-agent as it isn't much code and no one likes having features they use removed from a program. Do post-process hooks need to be added to rss2email or will it load them from a user owned directory? Can multiple post-process hooks be used? If both are possible, then maybe a post-process hook could replace user-agent.

Fun fact, https://github.com/rss2email/rss2email/pull/74/commits/f57513aa9868ce5d18148eb4eb787599c2b5d2fc#diff-86d7b95911fb37a51f57d39f3ab50678bdf8f85d9d6a1853bf05eb75729e8461L64 in #74 hijacked the static http user agent to add a customizable mail user agent.

amiryal:
> Fun fact, [f57513a#diff-86d7b95911fb37a51f57d39f3ab50678bdf8f85d9d6a1853bf05eb75729e8461L64](https://github.com/rss2email/rss2email/commit/f57513aa9868ce5d18148eb4eb787599c2b5d2fc#diff-86d7b95911fb37a51f57d39f3ab50678bdf8f85d9d6a1853bf05eb75729e8461L64) in #74 hijacked the static http user agent to add a customizable mail user agent.

Somehow I missed that! So, effectively, f57513aa9868ce5d18148eb4eb787599c2b5d2fc introduced an unintended change of behaviour in 3.11, which we are now trying to untangle. Anecdotally, that same commit also broke the `redirect` post process, which was fixed in 6e859aa49ad1842f86e7bcfc981a6e5d777ba063 (version 3.12), and another fix for its fallout was spread across 30ca438b2f5e620e316b221ba9e8881a762a910b and 745c1c41cd5aba49f61993ed9331da8ac0e29d57.

I would put it this way:

* A good netizen program identifies itself as the user-agent in all outgoing requests. (I.e. rss2email should not fall back to using feedparser’s identity.)
* In light of the above, f57513aa9868ce5d18148eb4eb787599c2b5d2fc introduced a bug in 3.11.
* To fix the bug, restore rss2email into the UA string and explain the effects clearly in the release notes.

Ignoring the discussion thus far, I hope we can agree on at least the first two bullets. The third bullet raises a couple of contention points:

* If users started to rely on feedparser in the UA, is it OK to break it for them? To that I would answer yes, it’s OK, because with fixing the bug I would like to bring our users back to good netizenship, and the release notes should explain that.
* If users started to rely on the fact that changing the mail header doesn’t affect the HTTP UA, is it OK to break it for them? To that I would answer yes, it’s OK, because that “feature” was never documented as such (documentation for the `user-agent` config was introduced in 3.11 together with the bug, never mentioning HTTP requests at all), and the release notes should explain that.

# TL;DR

If updating the manual accordingly and adding the support of some good release notes to explain the situation, I am still in favour of: 

> neither `http-user-agent` nor `mail-user-agent` should be added, and instead use the existing `user-agent` config for HTTP

amiryal:
> Do post-process hooks need to be added to rss2email or will it load them from a user owned directory?

The latter — they simply need to be somewhere in `sys.path`.

> Can multiple post-process hooks be used?

Only one can be named in the configuration, but there is nothing to stop users from calling multiple others in a chain from their custom post-process hook.

PR #210, opened 2022-03-06
Fix CHANGELOG for unreleased 1.14/2.0

amiryal:
Fixes https://github.com/rss2email/rss2email/pull/182#discussion_r820204939.


auouymous:
Good find.

Ekleog:
Thank you, I'd probably have forgotten about it at the time of release-cutting :) (btw feel free to let me know when you think I should cut a new release, while I don't have the right to add you as pypi owners you're following the development of rss2email much more closely than I have been doing so recently)

PR #213, opened 2022-03-16
Add http-header config option to pass custom HTTP headers.

rand0mdud3:
This can be used to pass cookies for examples for feeds that may require
them.

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -383,6 +383,16 @@ class Feed (object):
             kwargs['handlers'] = [
                 _urllib_request.ProxyHandler({ 'http': proxy, 'https': proxy })
             ]
+        extra_headers = {}
+        custom_headers = config['http-header']
+        if custom_headers:
+            for header in self.http_header.splitlines():
+                if ':' in header:
+                    key,value = header.split(':', 1)
+                    extra_headers[key.strip()] = value.strip()
+                else:
+                    _LOG.warning('malformed http-header: {}'.format(self.bonus_header))
+        kwargs['request_headers'] = extra_headers
         f = _util.TimeLimitedFunction('feed {}'.format(self.name), timeout, _feedparser.parse)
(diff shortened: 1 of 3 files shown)


rss2email/feed.py
```diff
@@ -383,6 +383,16 @@ def _fetch(self):
             kwargs['handlers'] = [
                 _urllib_request.ProxyHandler({ 'http': proxy, 'https': proxy })
             ]
+        extra_headers = {}
+        custom_headers = config['http-header']
+        if custom_headers:
+            for header in self.http_header.splitlines():
```
auouymous, line 389: Should `self.http_header` be `custom_headers`?

rss2email/feed.py
```diff
@@ -383,6 +383,16 @@ def _fetch(self):
...
+                if ':' in header:
+                    key,value = header.split(':', 1)
+                    extra_headers[key.strip()] = value.strip()
+                else:
+                    _LOG.warning('malformed http-header: {}'.format(self.bonus_header))
```
auouymous, line 394: Should `self.bonus_header` be `header`?

auouymous (changes requested):

rand0mdud3:
yes & yes... doh, should be more careful, thanks @auouymous !

auouymous:
It should probably be `http-headers` to indicate more than one header can be given. However, it doesn't overwrite all headers, so maybe something like `http-custom-headers`.

And it needs a test.

hanskuhn:
This feature would be helpful for consuming the weather.gov [API](https://www.weather.gov/documentation/services-web-api) as it requires 'Accept: application/atom+xml' and 'User-Agent:' headers.

mcepl:
Rebased version of the branch is available at https://github.com/openSUSE-Python/rss2email/tree/213-http-header-config

PR #52, opened 2019-05-11
allow setting user-agent value for fetching feeds

andresmrm:
Signed-off-by: Andres MRM <andres@inventati.org>

Related to: https://github.com/rss2email/rss2email/issues/48

PR #216, opened 2022-04-20
Fix a couple of HTTP user agent regressions

### LINKED ISSUE #120: user-agent setting has an unintuitive name: it sounds like HTTP user agent, but it's RFC5322 user agent

### LINKED ISSUE #154: option for change User-Agent in http headers

amiryal:
Based on work in and closes #209. Closes #120 and fixes #154 (ability to configure HTTP user agent). It also fixes a regression from version 3.11, where `feedparser` was used in the User-Agent for fetching feeds instead of the previously-used `rss2email`.

Read [this comment in the #209 thread](https://github.com/rss2email/rss2email/pull/209#issuecomment-1102715056) for the full context. The gist is in the commit message:

> It is worth noting that https://github.com/rss2email/rss2email/commit/f57513aa9868ce5d18148eb4eb787599c2b5d2fc (Use configured user-agent string and replace \_VERSION\_ with \_\_version\_\_ value, 2019-09-13) changed the semantics of the `user-agent` configuration unintentionally. This error was unfortunately added to the documentation in https://github.com/rss2email/rss2email/commit/574ba59ae7032e5d6e6a0ae77b22a06487f6c247 (Add user-agent documentation to man page, 2019-09-13), causing further confusion.

This PR is an attempt to clear up that confusion.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,9 @@ UNRELEASED
     * Fix exception when parsing a feed with a linebreak in its title
     * Add a new `subject-format` setting, customise the subject line
     * Removed '$' interpolation in config file to allow URLs containing dollar signs. Interpolation was not fully supported, and the placeholder would not survive a second save. Config files with ${...} placeholders will need to be manually edited after upgrading, or a save forced by adding and removing a fake feed before upgrading.
+    * Fix default HTTP User-Agent to use `rss2email` instead of `feedparser`, and fix `user-agent` setting to consistently apply to both outgoing emails and HTTP requests. This introduces a couple of potentially breaking changes:
+       o Configurations that came to rely on the `feedparser` UA string should adjust to the updated UA string going forward.
+       o Configurations that came to rely on the unintended behaviour, that configuring the User-Agent header in emails does not affect HTTP requests, should use a custom `post-process` hook instead.
 
 v3.13 (2021-04-03)
     * Switch to feedparser 6
--- a/r2e.1
+++ b/r2e.1
@@ -257,9 +257,6 @@ Wrap long lines at position. Any negative value for no wrapping, 0 for 78 width
 Select protocol from: sendmail, smtp, imap, maildir
 .IP sendmail
 Path to sendmail (or compatible)
-.IP user-agent
-String to use as User-Agent in outgoing emails. If present, __VERSION__ and
-__URL__ are replaced with rss2email version number and webpage
 .RE
 .SS SMTP configuration
 .IP smtp-auth
@@ -296,6 +293,10 @@ Path of maildir to write messages into
 Mailbox within maildir-path to write messages into
 .RE
 .SS Miscellaneous
+.IP user-agent
+String to use as HTTP User-Agent in web requests and as the User-Agent header in
+outgoing emails. If present, __VERSION__ and __URL__ are replaced with rss2email
+version number and webpage.
 .IP verbose
 Verbosity (one of 'error', 'warning', 'info', or 'debug').
 .RE
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -384,7 +384,7 @@ class Feed (object):
                 _urllib_request.ProxyHandler({ 'http': proxy, 'https': proxy })
             ]
         f = _util.TimeLimitedFunction('feed {}'.format(self.name), timeout, _feedparser.parse)
-        return f(self.url, self.etag, modified=self.modified, **kwargs)
+        return f(self.url, self.etag, modified=self.modified, agent=self.user_agent, **kwargs)
 
     def _process(self, parsed):
         _LOG.info('process {}'.format(self))
--- a/test/test.py
+++ b/test/test.py
@@ -192,6 +192,20 @@ def webserver_for_test_fetch(queue, num_requests, wait_time):
     finally:
         httpd.server_close()
 
+def webserver_for_test_user_agent(queue):
+    class AgentDumper(NoLogHandler):
+        def do_GET(self):
+            queue.put(self.headers["User-Agent"])
+            super().do_GET()
+
+    httpd = http.server.HTTPServer(('', 0), AgentDumper)
+    try:
+        port = httpd.server_address[1]
+        queue.put(port)
+        httpd.handle_request()
+    finally:
+        httpd.server_close()
+
 def webserver_for_test_if_fetch(queue, timeout):
     """Spawn a webserver for `timeout` seconds"""
     httpd = http.server.HTTPServer(('', 0), NoLogHandler)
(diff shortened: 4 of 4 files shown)


auouymous:
The patch looks fine and doesn't introduce a breaking change for anyone using the setting as an email user-agent or an http user-agent with the redirect hook. Anyone using it for email who needs to spoof a feed can just stick 'rss2email' somewhere in the spoofed user agent. And passing a spoofed browser user-agent in emails won't cause any issues. I'm fine with this if @Ekleog is.

Ekleog:
If I understand correctly, this changes user-agent to affect both HTTP and mail RFC822 user agent strings?

If so, I'm a bit sad in that it's technically a breaking change whereas my ideal world would have wanted an end-result of two orthogonal http-user-agent and one mail-user-agent configuration parameters; but you're both more involved in this software than I am and it actually fixes a real-life use case, so whatever and we can always introduce these two configuration parameters later on :)

The only thing I think could be made better would be splitting the two new changelog points to have a "breaking change" section, making it obvious to people that this is a change worthy of attention ?

amiryal:
> If I understand correctly, this changes user-agent to affect both HTTP and mail RFC822 user agent strings?

Nitpicky correction: yesterday I learned that the `User-Agent` mail header is not part of [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322), but rather a common but non-standard addition. It is documented as such in [an expired internet draft from 2013](https://datatracker.ietf.org/doc/html/draft-melnikov-email-user-agent-00).

To answer your question, yes, one config option to change both, hence moving the doc entry to the miscellaneous section. Effectively this PR reverts to pre- f57513aa9868ce5d18148eb4eb787599c2b5d2fc behaviour but with the added config option (which, as I understand it, is how #74 was intended in the first place!).

> The only thing I think could be made better would be splitting the two new changelog points to have a "breaking change" section, making it obvious to people that this is a change worthy of attention ?

Will do, thanks for the feedback.

auouymous:
There is nothing breaking about this change. The internal setting was always for email, feedparser and the redirect hook. #52 would have made it user-configurable and broken the redirect hook. #74 made it configurable and broke both feedparser and the redirect hook. I thought #74 had also implemented the user agent for emails, but that existed before, so it didn't hijack the user agent, it only broke it. #97 fixed the redirect hook, making it use the setting as an http agent once again. And #120 would have fixed feedparser in the same way. This current PR fully implements #120 to fix feedparser and restore original behavior of the internal setting.

We could add mail-user-agent and http-user-agent settings, and initialize them from user-agent. The user-agent setting could then be removed in a couple years, and anyone who updates during that time will have a seamless transition. The only reason to do this is so emails don't advertise with a spoofed agent.

amiryal:
```
$ git range-diff dd56bca...e1d8e21
1:  dd56bca ! 1:  e1d8e21 Fix a couple of HTTP user agent regressions
    @@ CHANGELOG: UNRELEASED
          * Fix exception when parsing a feed with a linebreak in its title
          * Add a new `subject-format` setting, customise the subject line
          * Removed '$' interpolation in config file to allow URLs containing dollar signs. Interpolation was not fully supported, and the placeholder would not survive a second save. Config files with ${...} placeholders will need to be manually edited after upgrading, or a save forced by adding and removing a fake feed before upgrading.
    -+    * Fix unintended use of `feedparser` User-Agent string in outgoing HTTP requests. Configurations that came to rely on the `feedparser` UA string should adjust to the updated UA string going forward.
    -+    * Fix regression in `user-agent` configuration option to affect outgoing HTTP requests. Configurations that came to rely on the unintended behaviour, that configuring the User-Agent header in emails does not affect HTTP requests, should use a custom `post-process` hook instead.
    ++    * Fix default HTTP User-Agent to use `rss2email` instead of `feedparser`, and fix `user-agent` setting to consistently apply to both. This introduces a couple of potentially breaking changes:
    ++       o Configurations that came to rely on the `feedparser` UA string should adjust to the updated UA string going forward.
    ++       o Configurations that came to rely on the unintended behaviour, that configuring the User-Agent header in emails does not affect HTTP requests, should use a custom `post-process` hook instead.
      
      v3.13 (2021-04-03)
          * Switch to feedparser 6

**Comment by amiryal** [comment]
@auouymous:
> We could add mail-user-agent […]. The only reason to do this is so emails don't advertise with a spoofed agent.

If you ask me, I would avoid overloading the term user-agent, and leave it to the more common HTTP usage. If anyone ever offers to resolve the issue of emails advertising a spoofed agent, then my suggestions would be:
1. Use a separate new string, i.e. do not use the `user-agent` setting at all for emails.
2. If adding a configuration option for it (why?), then give it a completely different name (like `mailer-identification`, alluding to `X-Mailer`).

File: CHANGELOG [hunk 1]
```diff
@@ -3,6 +3,9 @@ UNRELEASED
     * Fix exception when parsing a feed with a linebreak in its title
     * Add a new `subject-format` setting, customise the subject line
     * Removed '$' interpolation in config file to allow URLs containing dollar signs. Interpolation was not fully supported, and the placeholder would not survive a second save. Config files with ${...} placeholders will need to be manually edited after upgrading, or a save forced by adding and removing a fake feed before upgrading.
+    * Fix default HTTP User-Agent to use `rss2email` instead of `feedparser`, and fix `user-agent` setting to consistently apply to both. This introduces a couple of potentially breaking changes:
```
auouymous, line 6: `... consistently apply to both outgoing emails and HTTP requests.` would make more sense because there is no context to suggest what `both` is.

auouymous (changes requested):

auouymous (approved):

auouymous:
I am fine with this PR simply reverting r2e to its previous behavior. Nothing should break, other than advertising the user-agent to feed servers, which is most likely just "rss2email ...". It seems like a bug fix to me.

MUA is Mail User Agent, so `mail-user-agent` makes the most sense if separate options, with `http-user-agent` for web requests, just to make it clear. Using `mailer-identification` or `mailer-agent` implies r2e sends `X-Mailer`, which it doesn't. Should we also send `X-Mailer`, maybe. Should we not send `User-Agent`, no because that would be a breaking change for anyone checking for it. Having separate settings would be ideal for anyone who wants to spoof the HTTP user agent **and** filter emails sent by r2e. Could they use a post-process hook to change User-Agent or add X-Mailer, sure, but that is no where near as easy as setting something called `mail-user-agent`.

amiryal:
> MUA is Mail User Agent […]

I agree, FWIW.

Issue #194, opened 2021-07-19
Single-part digest?


squeakbat:
Not all e-mail clients (ahem, mail user agents) display the multipart digest format in a usable way, for example k-9 on android and gmail just show the parts as downloadable attachments.

It would be nice to have an optional digest format as a single (html or plaintext) body.  It can basically look like the concatenation of the digest parts with some separator.  Alternatively (the cheap and nasty version), we can just make the message multipart/mixed wrapped around text/html, instead of multipart/digest around message/rfc822 around text/html.

What do y'all think?

Since I want this feature, I'll be happy to write the code, obvs.

devilgate:
I completely agree, and this is either a duplicate of #173, or at least closely related. I've started looking at that one a couple of times, but never quite got up the energy to work on it properly, so if you do make this change, I'd like to know.

squeakbat:
Sorry about the delay in answering.  (It came in when I had appendicitis (no kidding), then it got buried.)

It's a messy issue.  I suppose we can add an option like "digest-one-part" and keep the old format as default.  But given how few mail user agents can display multipart/digest message/rfc822, I'm really more inclined to just switch to a single-part digest format as the only option.  It'll break compatibility with digest-post-process though.

The code change to support both formats is a little complicated, but doable (obviously).  I think mostly feed._end_digest() will be more complicated.

I would like some feedback on this from maintainers before going forward.  Thanks!

kriswilk:
IMHO, this issue needs urgent action. rss2email's current digest format is objectively useless in gmail (and others by the sound if it). I appreciate wanting to be standards-compliant. But it's more important to be functionally useful. If that means reverting to a single-part digest, then so be it.

![image](https://user-images.githubusercontent.com/382436/175044936-690b4ec0-3e86-4e35-be89-2a56a099dc9f.png)

xrat:
I am afraid it won't help anyone here but just in case: I wrote an Awk script that flattens r2e's multipart/digest: https://gist.github.com/xrat/d8b92e32e22af0cf8e05c1e7538cd2f5

auouymous:
Would it be possible to translate that awk code to python in a post-processing script?

squeakbat:
This is my version.  There's a bonus postproc function for adding feed name to the subject (for nondigest messages).

But this is only a stopgap.  We should do it right in the first place.

```
import email.mime.multipart as mp
import email.header as h

# prepend feed name to subject
[...]

PR #221, opened 2022-06-23
Add optional multipart/mixed digest format.

squeakbat:
The previous format (multipart/digest), was incompatible with many
popular e-mail readers (such as gmail), which made it essentially
unusable.

The new format is multipart/mixed, with each part consisting
of a single non-digest message (of type text/plain or text/html).
One possible incorrectness is that the parts still retain message
headers (subject, date, etc.) that are not really used by normal text
attachments.

This is an incompatible change.  It looks different, and may break
existing digest-post-proc functions.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,5 @@
 UNRELEASED
+    * New `digest-type` configuration adds optional more widely supported `multipart/mixed` format
     * New argument `--only-new` on the `add` command to ignore entries in feed
       when added, so only new entries will be sent.
     * Fix crash on html-mail entries with no URL


squeakbat:
Would this be more palatable as an option and/or as a straight one-part message?

The problem with a default-off option is that digest will still be practically useless by default.  The problem with one-part message format is it'll be even less compatible with existing postproc filters.

amiryal:
I like the idea of a utility boost, but I hesitate to accept backwards-incompatible changes without a mandate. This leads me to think that this project needs to prepare for a major version update, where changes such as this and others like #216 can be accepted without hesitance.

Another way to go about it is to develop a `digest-post-process` hook that converts the given `multipart/digest` into a `multipart/mixed` (while somehow satisfying the code in `_send_digest` that extracts the date). To make it usable and discoverable, it may be a good idea to add a section in the manual about it (and if feeling generous, also briefly describe all the other hooks from the `rss2email.post_process` package in the same section). Interestingly, this idea can go both ways, i.e., should we choose to change the default to `multipart/mixed`, then including such a hook to convert back to the old `multipart/digest` would be handy, for those who still depend on it.

squeakbat:
I posted such a digest-post-process in the discussion: https://github.com/rss2email/rss2email/issues/194#issuecomment-1163818487.  It preserves the nesting level with an additional multipart/mixed wrapping.  Though it seems wrong, as a permanent solution, to force the user to resort to a postprocessor (even a well-documented one).

I'm willing to make the built-in feature optional and default off.  (Mostly, I can't come up with a good name for that config option.  "multipart-mixed-digest" sounds so lame.)  This seems to be a good compromise.  There is no incompatibility, because if a user sets that option, then they are aware that any digest-post-process filter they write has to handle the multipart/mixed format.

As a separate issue, is single-part digest better then multipart/mixed?  It's cleaner, but lacks the structure that can help guide postproc filters.

amiryal:
Yesterday I read [a report about an interview with Linus Torvalds](https://lwn.net/Articles/899182/). The following quote reminded me of this discussion:

> Torvalds is not just a developer, he is also a user, and he thinks that the most annoying problem that users experience is "doing a software upgrade and things stop working". He cannot change all of the other software projects out there, which generally have different policies, but he has been "very hard-nosed" with kernel policy to try to ensure that programs continue to run after a upgrade.

I completely agree with the sentiment, that your old setup should just work, following an upgrade.

#

>  There is no incompatibility, because if a user sets that option, then they are aware that any digest-post-process filter they write has to handle the multipart/mixed format.

True, it may be fine from the user’s perspective, but think about it from a software maintainability standpoint. I do not want `if…else…` statements to be sprinkled all over.

One way to work around the naming issue is to leave the option name as `digest`, but make it a string instead of boolean, and restrict it to the following options:
* `multipart/digest`
* `multipart/mixed` → the new implementation
* any value that previously evaluated to `True` → alias for `multipart/digest`
* any value that previously evaluated to `False` → the default, not a digest

#

> As a separate issue, is single-part digest better then multipart/mixed? It's cleaner, but lacks the structure that can help guide postproc filters.

I don’t think any format is better than the other. So long as we keep a multipart default, users may postprocess further to their liking. A single-part digest is yet another example filter that we can provide and document.

squeakbat:
I do agree with the sentiment.  I wouldn't have suggested breaking compatibility except that I don't think anybody is really using digests.

I also like keeping the `digest` option and giving it new values.  I don't think it keeps code paths simpler (however many kinds of output we support is how complicated it is), but it does heads off further proliferation of options.

It's not all that straightforward to flatten a multipart/mixed message after the fact, because each part is a full html document, but it can be done.

squeakbat:
Oh yeah, anyway, I'll give it a go.

squeakbat:
OK.  After looking at the code and thinking about it, I think it much better to have a separate config option "digest-type" that defaults "multipart/digest".  It's conceptually unclean to mix boolean and string values (even as an compatibility kludge) and the code really doesn't like to do it.

squeakbat:
OK sorry I made the mistake of pulling from before making this change.  So, if it's better I can trash this one, make a clean branch, and submit a new pull request with a clean, single commit.

squeakbat:
Thanks!

The commit message will be pretty ugly and hard to follow, with "change digest format..."  followed by "revert ...", with merges from master in the middle.  But I guess I'm okay with that.

squeakbat:
Oh yeah, I noticed I have a typo in a comment, two slashes: multipart//mixed.

amiryal:
I fixed the typo, removed a pair of redundant parentheses, and added an entry to the CHANGELOG.

```diff
diff --git a/CHANGELOG b/CHANGELOG
index 2dbcf78..c245546 100644
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,5 @@
 UNRELEASED
+    * New `digest-type` configuration adds optional more widely supported `multipart/mixed` format
     * New argument `--only-new` on the `add` command to ignore entries in feed
       when added, so only new entries will be sent.
     * Fix crash on html-mail entries with no URL
diff --git a/rss2email/feed.py b/rss2email/feed.py
index b4d62d4..df21d85 100644
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -1005,12 +1005,12 @@ class Feed (object):
         return digest
 
     def _append_to_digest(self, digest, message):
-        if (self.digest_type == 'multipart/digest'):
+        if self.digest_type == 'multipart/digest':
             part = _MIMEMessage(message)
             part.add_header('Content-Disposition', 'attachment')
             digest.attach(part)
         else:
-            # multipart//mixed
+            # multipart/mixed
             digest.attach(message)
 
     def _send_digest(self, digest, sender):
```

squeakbat:
Thank you!

Yeah, I tried to write that conditional in C there.

rss2email/feed.py
```diff
@@ -1011,8 +1021,12 @@ def _send_digest(self, digest, sender):
         any messages in the digest, don't call this function.
         """
         digest['From'] = sender
-        last_part = digest.get_payload()[-1]
-        last_message = last_part.get_payload()[0]
+        if (self.digest_type == 'multipart/digest'):
```
auouymous, line 1024: More parentheses.

auouymous (changes requested):

rss2email/feed.py
squeakbat: Should have realized it.  They were copies of each other.

auouymous (approved):

amiryal:
Thanks, @squeakbat!

Amir (usually yxejamir in #rss2email on libera.chat)

squeakbat:
Thank you all for the help!

Issue #206, opened 2022-02-14
pypip.in badges are broken in README


lacasseio:
New to the project but though about mentioning the pypip.in badge are broken:

<img width="714" alt="Screen Shot 2022-02-14 at 8 30 11 AM" src="https://user-images.githubusercontent.com/22181740/153873257-f3ab5938-6cc2-4c7a-806b-774a68a5bbe5.png">

I dug a little bit but could only find issues around 2015 where [the domain wasn't renewed](https://github.com/badges/pypipins/issues/37). However, there are [still issues open about the broken badges](https://github.com/badges/pypipins/issues/39). Some linked PR seems to indicate other repositories simply removed the badges.

amiryal:
> Some linked PR seems to indicate other repositories simply removed the badges.

Hint, hint. ☺

auouymous:
Fixed by #222.

PR #222, opened 2022-07-27
Fix README badges

polm:
This fixes the badges at the top of the README, which were reported broken in #206.


polm:
You can preview this [here](https://github.com/polm/rss2email/tree/fix/readme-badges).

auouymous:
Thank you for the fix.

PR #224, opened 2022-10-02
Document same-server-fetch-interval

polyzen:
https://github.com/rss2email/rss2email/pull/60

### DIFF
--- a/r2e.1
+++ b/r2e.1
@@ -185,6 +185,8 @@ Set this to default To email addresses.
 Set an HTTP proxy (e.g. 'http://your.proxy.here:8080/')
 .IP feed-timeout
 Set the timeout (in seconds) for feed server response
+.IP same-server-fetch-interval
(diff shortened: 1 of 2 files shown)


r2e.1
```diff
@@ -185,6 +185,8 @@ Set this to default To email addresses.
 Set an HTTP proxy (e.g. 'http://your.proxy.here:8080/')
 .IP feed-timeout
 Set the timeout (in seconds) for feed server response
+.IP same-server-fetch-interval
+Specify a sleep interval between consecutive fetches from the same server
```
polyzen, line 189:
```suggestion
Specify a sleep interval (in seconds) between consecutive fetches from the same server
```

auouymous:
The comment in config.py should be changed to match man page, as it does not mention the fetches are "consecutive".

polyzen:
They now match. 👍

auouymous:
Thank you.

PR #230, opened 2022-11-04
HTTP statuses cleanup

### LINKED ISSUE #229: 303 HTTP status issue
When using rss2email on https://www.nature.com/nphys.rss I'm getting HTTP 303 errors. Feedparser seems to handle getting the feed content OK, but feed.py raises an error. I'm wondering if this [line](https://github.com/rss2email/rss2email/blob/7c1a2e1d705661c2de0982a5c4c7e6b0e657bf1e/rss2email/feed.py#L414)  should be changed to also exclude 303 status.

Happy to create a PR if needed.

amiryal:
Fixes #229 by not attempting to enumerate all non-error statuses explicitly.

While researching for this fix, I noticed that HTTP status 308 (Moved Permanently) is another form of permanent redirect and corrected the code to include that case. Note that currently feedparser does not handle HTTP status 308 as a redirect, but this will change when (hopefully) https://github.com/kurtmckee/feedparser/pull/326 gets merged and released there.

In addition, while I was in the area, I added handling of status 410 (Gone) in similar vein to the handling of permanent redirects, by marking the feed as inactive (as suggested [in the feedparser documentation](https://feedparser.readthedocs.io/en/latest/http-redirect.html#noticing-feeds-marked-gone)).

### DIFF
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -411,6 +411,10 @@ class Feed (object):
             _LOG.info('redirect {} from {} to {}'.format(
                     self.name, self.url, parsed['url']))
             self.url = parsed['url']
+        elif status == 410:
+            _LOG.info('deactivate {} because {} is gone'.format(
+                    self.name, self.url))
+            self.active = False
         elif status >= 400:
             raise _error.HTTPError(status=status, feed=self)
 
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -407,11 +407,11 @@ class Feed (object):
         warned = False
         status = getattr(parsed, 'status', 200)
         _LOG.debug('HTTP status {}'.format(status))
-        if status == 301:
+        if status in [301, 308]:
             _LOG.info('redirect {} from {} to {}'.format(
                     self.name, self.url, parsed['url']))
             self.url = parsed['url']
-        elif status not in [200, 302, 304, 307]:
+        elif status >= 400:
             raise _error.HTTPError(status=status, feed=self)
 
         http_headers = parsed.get('headers', {})


auouymous:
The PR looks good but r2e doesn't appear to save the `url` or `active` changes made by these statuses. Do either of these work for you?

```
r2e add test301 http://borg.qzx.com/test-rss.php?test=301
r2e -VVVV run test301 --no-send 2>&1 |grep redirect

r2e add test410 http://borg.qzx.com/test-rss.php?test=410
r2e -VVVV run test410 --no-send 2>&1 |grep deactivate
```

arunpersaud:
Just wanted to check in what the status is. Would be great to get this merged (I currently have to patch this for rss2email to work on my computer). Thanks!

amiryal:
> Do either of these work for you?
> […]

I didn’t try, but I trust you that they don’t.

@auouymous: any objection to merging this despite the seemingly useless code paths? If anyone cares, it can still be cleaned up afterwards.

auouymous:
I added some comments in the code https://github.com/rss2email/rss2email/commit/2f8274bb7969d67f3e1423266a0d3d01b6d43755. I believe that is the only way to save these changes.

auouymous:
@amiryal While looking at #240, I found the new 410 also logs a bunch of errors, because it doesn't return or raise to avoid the checks. Changing to a warning, and returning seems to produce the cleanest log output. Do you agree with this change?

```patch
--- a/rss2email/feed.py
+++ b/rss2email/feed.py
@@ -413,10 +413,11 @@ class Feed (object):
             self.url = parsed['url']
             # TODO: `url` is not saved -- add config option to call feeds.save_config() in run command
         elif status == 410:
-            _LOG.info('deactivate {} because {} is gone'.format(
+            _LOG.warning('deactivate {} because {} is gone'.format(
                     self.name, self.url))
             self.active = False
             # TODO: `active` is not saved -- add config option to call feeds.save_config() in run command
+            return
         elif status >= 400:
             raise _error.HTTPError(status=status, feed=self)
```

amiryal:
@auouymous 
> Do you agree with this change?

Yes, LGTM.

Issue #218, opened 2022-06-13
Official docker image?


kriswilk:
Are there plans for an official rss2email docker image? It would be convenient to manage rss2email as I do other dockerized apps (linkding, node-red, miniflux, ...).

It looks like @rgibert began in #207 but then closed it. Is there any interest in doing this? I'd be willing to help as much as my experience with rss2email and docker allow...

rgibert:
I accidentally opened that PR before it was ready, I got side-tracked with some other work but intend to work on getting this ready for a proper PR in the future.

PR #235, opened 2023-03-29
Add docker image

plantroon:
Apparently there was a similar PR/suggestion in the past but it was withdrawn. I am reviving that effort.

I wrote a minimal distroless (there's only python) and rootless Dockerfile. I successfully use this in my more elaborate setup with rssbridge and chadburn scheduling. The docker-compose example in the README.rst is taken from my setup.

This PR contains a github automation for building a docker image for ghcr.io.

I will implement any necessary changes if requested.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,96 @@
+name: Docker
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+on:
+  schedule:
+    - cron: '43 23 * * *'
+  push:
+    branches: [ "master" ]
+    # Publish semver tags as releases.
+    tags: [ 'v*.*.*' ]
+  pull_request:
+    branches: [ "master" ]
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
+  IMAGE_NAME: ${{ github.repository }}
+
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      # This is used to complete the identity challenge
+      # with sigstore/fulcio when running outside of PRs.
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Install the cosign tool except on PR
+      # https://github.com/sigstore/cosign-installer
+      - name: Install cosign
+        if: github.event_name != 'pull_request'
+        uses: sigstore/cosign-installer@f3c664df7af409cb4873aa5068053ba9d61a57b6 #v2.6.0
+        with:
+          cosign-release: 'v1.13.1'
+
+
+      # Workaround: https://github.com/docker/build-push-action/issues/461
+      - name: Setup Docker buildx
+        uses: docker/setup-buildx-action@79abd3f86f79a9d68a23c75a09a9a85889262adf
+
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Extract metadata (tags, labels) for Docker
+      # https://github.com/docker/metadata-action
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      # Build and push Docker image with Buildx (don't push on PR)
+      # https://github.com/docker/build-push-action
+      - name: Build and push Docker image
+        id: build-and-push
+        uses: docker/build-push-action@ac9327eae2b366085ac7f6a2d02df8aa8ead720a
+        with:
+          context: .
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+
+      # Sign the resulting Docker image digest except on PRs.
+      # This will only write to the public Rekor transparency log when the Docker
+      # repository is public to avoid leaking data.  If you would like to publish
+      # transparency data even for private images, pass --force to cosign below.
+      # https://github.com/sigstore/cosign
+      - name: Sign the published Docker image
+        if: ${{ github.event_name != 'pull_request' }}
+        env:
+          COSIGN_EXPERIMENTAL: "true"
+        # This step uses the identity token to provision an ephemeral certificate
+        # against the sigstore community Fulcio instance.
+        run: echo "${{ steps.meta.outputs.tags }}" | xargs -I {} cosign sign {}@${{ steps.build-and-push.outputs.digest }}
--- a/README.rst
+++ b/README.rst
@@ -46,11 +46,11 @@ __ `Ubuntu package`_
 Docker
 ------
 
-A docker package exists for this project: 
+A docker package exists for this project::
 
   $ git pull ghcr.io/rss2email/rss2email:master
 
-docker-compose.yaml
+docker-compose.yaml::yaml
 ~~~~~~~~~~~~~~~~~~~
   version: '2'
   services:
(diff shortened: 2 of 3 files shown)


plantroon:
This would close #218 for example

Can anyone approve this please? :) I am happy to work on any adjustments as requested.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,96 @@
+name: Docker
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+on:
+  schedule:
+    - cron: '43 23 * * *'
```
auouymous, line 10: Does this build an image every day, even if no changes have been made?

README.rst
```diff
@@ -43,6 +43,37 @@ __ `OpenBSD package`_
...
+  services:
+    rss2email:
+      image: ghcr.io/rss2email/rss2email:master
+      volumes:
+        - ./config:/config
+        - ./data:/data
+      entrypoint: "/venv/bin/python3 -c 'import threading; threading.Event().wait()'"
+      labels:
+        chadburn.enabled: "true"
+        chadburn.job-exec.rss2email.schedule: "@every 5m"
```
auouymous, line 75: We shouldn't be suggesting users run r2e more than once per day.

README.rst
```diff
@@ -53,6 +53,8 @@ A docker package exists for this project::
 Then it can be used like::
 
   $ docker run  -v ./data:/data -v ./config:/config rss2email list
+
```
auouymous, line 56: Whitespace on blank line.

README.rst
```diff
@@ -65,9 +72,11 @@ docker-compose.yaml::yaml
         chadburn.job-exec.rss2email.schedule: "@every 5m"
         chadburn.job-exec.rss2email.command: "r2e run"
 
-The example docker-compose snippet sets up chadburn (or ofelia) job scheduler to run r2e periodically. For the labels to be picked up, it is required that the container keeps running which is ensured with the entrypoint. Python's Event().wait() was used as there is no shell and this method should use less cpu than sleeping or looping (with sleep). You can of course abandon this and just call r2e from any other job scheduler as this entrypoint only circumvents chadburn/ofelia limitations.
+The example docker-compose snippet sets up chadburn (or ofelia) job scheduler to run r2e periodically.
```
auouymous, line 78: End of line whitespace.

README.rst
```diff
@@ -46,11 +46,11 @@ __ `Ubuntu package`_
 Docker
 ------
 
-A docker package exists for this project: 
+A docker package exists for this project::
```
auouymous, line 49: Why the double colons here and in other places?

README.rst
```diff
@@ -65,9 +72,11 @@ docker-compose.yaml::yaml
...
+The example docker-compose snippet sets up chadburn (or ofelia) job scheduler to run r2e periodically. 
+
+Some background for entrypoint override: For the labels to be picked up, it is required that the container keeps running which is ensured with the entrypoint. Python's Event().wait() was used as there is no shell and this method should use less cpu than sleeping or looping (with sleep). You can of course abandon this and just call r2e from any other job scheduler as this entrypoint only circumvents chadburn/ofelia limitations.
 
-You need to run `chown 65532:65532 config data` to have the correct ownership of files for this rootless docker image.
+You need to run ``chown 65532:65532 config data`` to have the correct ownership of files for this rootless docker image.
```
auouymous, line 82: Why the double quotes here and in other places?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,96 @@
...
+  pull_request:
+    branches: [ "master" ]
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
+  IMAGE_NAME: ${{ github.repository }}
+
+
```
auouymous, line 24: There are multiple instances of double line spacing in this file.

auouymous (changes requested):
@amiryal Do you know anything about docker?

README.rst
amiryal: @auouymous 
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

README.rst
amiryal: @auouymous 
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

README.rst
amiryal: @plantroon 
The guide should either walk the user through the entire setup (e.g., how does chadburn get installed and invoked?) or leave out invocation details for users to figure out for themselves.

README.rst
plantroon: My first time using RST - I saw this used many times in this README,  isn't this how one is supposed to do a codeblock?

README.rst
plantroon: Saw it in other places in the README used to do an inline literal. So same as the above, this was my first time using RST, so I am not sure how better to do it

README.rst
plantroon: Sorry, will fix

README.rst
plantroon: agreed, I will bring this in line with cron example

.github/workflows/docker-publish.yml
plantroon: I will bring this in line with the other python workflow, to run on push. Yes the schedule was stupid and I forgot to remove it from here.

.github/workflows/docker-publish.yml
plantroon: I'll bring this more in line with the other workflow too, the comments are also possibly useless and come from some other example. The actions are descriptive enough by themselves.

README.rst
plantroon: Will fix. Got a better setup now so that I won't miss these anymore

plantroon:
I'll squash commits when I fix this so that it doesn't mess up history too much

README.rst
```diff
@@ -43,6 +43,49 @@ __ `OpenBSD package`_
...
+      entrypoint: "/venv/bin/python3 -c 'import threading; threading.Event().wait()'"
+      labels:
+        chadburn.enabled: "true"
+        chadburn.job-exec.rss2email.no-overlap: "true"
+        chadburn.job-exec.rss2email.schedule: "@every 5m"
+        chadburn.job-exec.rss2email.command: "r2e run"
+
+The example docker-compose snippet sets up chadburn (or ofelia) job scheduler to run r2e periodically. 
+
+Some background for entrypoint override: For the labels to be picked up, it is required that the container keeps running which is ensured with the entrypoint. Python's Event().wait() was used as there is no shell and this method should use less cpu than sleeping or looping (with sleep). You can of course abandon this and just call r2e from any other job scheduler as this entrypoint only circumvents chadburn/ofelia limitations.
```
amiryal, line 80: This seems like too much detail for something that is outside of r2e. In future iterations of chadburn this workaround will be made redundant, but it will remain here until someone cares enough to update. Also, it seems a little excessive for something to be kept running indefinitely just to wake up a couple of times per day. The CPU usage while running may be close to 0%, but RAM is also a resource, and some cloud platforms might charge you for every second of run time.

README.rst
```diff
@@ -43,6 +43,49 @@ __ `OpenBSD package`_
...
+Here is an example docker-compose.yaml::
+
+  version: '2'
+  services:
+    rss2email:
+      image: ghcr.io/rss2email/rss2email:master
+      volumes:
+        - ./config:/config
+        - ./data:/data
+      entrypoint: "/venv/bin/python3 -c 'import threading; threading.Event().wait()'"
```
amiryal, line 71: @plantroon 
A later paragraph mentions there is no shell in the container, but this looks like a shell command line to me. Educate me here — does docker-compose parse this line and call `exec` with the parsed parameters?

@auouymous 
To answer your opening question, no, I do not have much hands-on experience with Docker.

plantroon: @amiryal yes, this is executed directly by Docker (no need for compose, specifying entrypoint in docker run would yield the same result) even without a shell available in the container.

The rationale behind this is for it to keep running as a service so that a job scheduler can pick it up and exec `r2e run`, which to my understanding runs the same way as when you would call `docker exec`.

README.rst
plantroon: The CPU usage here should be lower than any sleep command available.

I don't think it's possible for Chadburn or Ofelia to do exec on a stopped service, though other job schedulers operating outside of Docker might be able to.

I originally had this set up to run via docker run in cron, but opted to go for a cleaner docker-centric setup where I have everything in one place.

README.rst
auouymous: Oops, I thought it was markdown.

README.rst
plantroon: @amiryal  I understand the confusion now. I specified it in the shell command form, therefore it runs through a shell - and even distroless images have `sh` binary in them by default, but I thought they didn't.

I will change this to use the exec command form to make it cleaner.

README.rst
plantroon: In future iterations of chadburn this workaround will be made redundant - is this related to the job-run functionality? I was not aware of that coming :)

plantroon: @amiryal In future iterations of chadburn this workaround will be made redundant - is this related to the job-run functionality? I was not aware of that coming :)

plantroon (commented):
Fixed all your concerns. I additionally requested @amiryal  to take another look at the chadburn part.

Centro1993:
Great Stuff, would love to see this added!

PR #237, opened 2023-04-05
fix: add support for email threading per RSS feed

monperrus:
in email clients, emails with same `References` header are grouped together.

### DIFF
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -54,6 +54,7 @@ import subprocess as _subprocess
 import sys as _sys
 import time as _time
 import os as _os
+import hashlib
 
 import html2text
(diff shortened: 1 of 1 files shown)


amiryal:
Sorry, but this feature does not look desirable. It only makes sense to thread together replies to the same OP, but usually RSS feed items represent a single OP each (with no replies). Furthermore, RSS feeds that _do_ represent replies to a single OP usually indicate each reply’s separate author name, hence making your implementation incorrect.

If your use case requires it, then you can look into implementing it privately as a [post-process hook](https://github.com/rss2email/rss2email/blob/master/r2e.1#L226).

PR #241, opened 2023-09-02
Support sending mail via LMTP

pks-t:
When rss2email runs on the same host as the target mail server, then it is quite common that the server exposes a locally accessible LMTP socket. The LMTP protocol is quite similar to the SMTP protocol, but generally has no mail queue and thus requires no MDA. Instead, the mail is directly porcessed by the mail server without any intermediates.

Implement LMTP support such that it becomes possible to deliver mail to such servers directly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -205,6 +205,37 @@ your login credentials to the server.
 
 __ `Simple Mail Transport Protocol`_
 
+LMTP
+----
+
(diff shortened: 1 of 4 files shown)


auouymous (approved):

auouymous:
Thank you for the enhancement.

PR #243, opened 2023-10-06
Feed Filters

beastwick:
Allow the user to configure filters on their feeds, so they will only be emailed items in which they are only interested. This can be accomplished by setting the new 'filters' field in the feeds options found in the rss2email.cfg file. The value can be a '|' delimited list of values.

Within rss2email.cfg, the user must manually add this field. Example:

```
[feed.kernel]                                                                                                                                                     
url = https://www.kernel.org/feeds/kdist.xml                                                                                                                      
filters = stable                                                                                                                                                                                                                                                                                                                    

[feed.debian.linux]                                                                                                                                               
url = https://tracker.debian.org/pkg/linux/rss                                                                                                                    
filters = testing|unstable
```


auouymous:
Good idea!

It might be better to use something like "filter_subject" so users know it only matches the subject.

Does it print a message for feeds without a filter?

Since it uses regular expression matching, there is no reason to split and iterate, just pass the entire filter and perform a single match. Your examples will still work.

gPodder has a filter extension with block and except options, so it can block all with an empty field, and make exceptions with the except field. That gives the user more choice in what to block or allow. It also has checkboxes for each field to ignore case and switch between text matching and regular expressions. That would be more difficult to specify in the config, but might help users who don't know how to use regular expressions. We could ignore case for all filters, but that can cause issues for certain words.

beastwick:
Hey @auouymous thanks for your feedback. I updated the variable name to filter_subject as suggested, removed the pipe delimiter logic as it is indeed a basic regex, and added some logging to stdout.

PR #245, opened 2023-10-07
feat: align Python versions from 3.8 to 3.12

marcaurele:
Add CI validation for Python 3.12 and align the Python version information on Pypi to reflect the current list supported.


auouymous:
Thanks for the update.

PR #248, opened 2023-12-12
quote-printable encoding of the From header

monperrus:


amiryal:
On which version of Python did you observe the problem?

```
Python 3.11.6 (main, Oct  8 2023, 05:06:43) [GCC 13.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.13.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from email.utils import formataddr as _formataddr

In [2]: _formataddr(('äṁïṙ', 'amir@example.com'))
Out[2]: '=?utf-8?b?w6ThuYHDr+G5mQ==?= <amir@example.com>'
```
This ↑ seems to already be the desired behaviour, is it not?

amiryal:
In a comment in the code you mention MIMETEXT:

> encode the name, otherwise MIMETEXT will later encode the whole From header

What is it, and how can you demonstrate that it’s doing anything wrong?

As can be seen [in the test data](https://github.com/rss2email/rss2email/blob/adc16fd2d014b016b8033e07d62a5e657a213f33/test/data/bbc-chinese/1.expected#L5), the `From` header is already encoded correctly, with the name part in a “RFC 2047 chrome” and the address part as untouched ASCII.

monperrus:
Agree! I probably configured something wrong, because it seems to work now. Thanks for the great tool!

PR #252, opened 2024-01-13
support oauth2 login

arunpersaud:
Initial support for oauth2 login (at least for gmail).
The user needs to set 4 variables in the config file:

imap-oauth2 is a boolean that needs to be set to True. 
imap-clientid, imap-clientsecret, imap-refreshtoken are needed to log in.

The new file oauth2.py also includes code to generate a refresh token. Currently the function needs to be called manually though and this should probably be integrated into r2e before merging.

I wanted to create this PR to start a discussion on what else needs to change so that this can be integrated in the future.

I'm running a similar code (but currently not the same) on my desktop where this is working already.

Looking forward to hearing back from you.

### DIFF
--- a/rss2email/email.py
+++ b/rss2email/email.py
@@ -1,6 +1,6 @@
 # -*- encoding: utf-8 -*-
 #
-# Copyright (C) 2012-2022 Amir Yalon <git@please.nospammail.net>
+# Copyright (C) 2012-2024 Amir Yalon <git@please.nospammail.net>
 #                         Arun Persaud <apersaud@lbl.gov>
 #                         Dmitry Bogatov <KAction@gnu.org>
 #                         George Saunders <georgesaunders@gmail.com>
@@ -254,18 +254,33 @@ def lmtp_send(recipient, message, config=None, section='DEFAULT'):
 def imap_send(message, config=None, section='DEFAULT'):
     if config is None:
         config = _config.CONFIG
-    server = config.get(section, 'imap-server')
-    port = config.getint(section, 'imap-port')
-    _LOG.debug('sending message to {}:{}'.format(server, port))
-    ssl = config.getboolean(section, 'imap-ssl')
-    if ssl:
+    server = config.get(section, "imap-server")
+    port = config.getint(section, "imap-port")
+    _LOG.debug("sending message to {}:{}".format(server, port))
+    ssl = config.getboolean(section, "imap-ssl")
+    oauth2 = config.getboolean(section, "imap-oauth2")
+    if oauth2:
+        imap = _imaplib.IMAP4_SSL(server, ssl_context=_ssl.create_default_context())
+    elif ssl:
         imap = _imaplib.IMAP4_SSL(server, port)
     else:
         imap = _imaplib.IMAP4(server, port)
     try:
-        if config.getboolean(section, 'imap-auth'):
-            username = config.get(section, 'imap-username')
-            password = config.get(section, 'imap-password')
+        if oauth2:
+            username = config.get(section, "imap-username")
+            client_id = config.get(section, "imap-clientid")
+            client_secret = config.get(section, "imap-clientsecret")
+            refresh_token = config.get(section, "imap-refreshtoken")
+            auth_string = _oauth2.get_oauth2_auth_string(
+                client_id,
+                client_secret,
+                refresh_token,
+                username,
+            )
+            imap.authenticate("XOAUTH2", lambda x: auth_string)
+        elif config.getboolean(section, "imap-auth"):
+            username = config.get(section, "imap-username")
+            password = config.get(section, "imap-password")
             try:
                 if not ssl:
                     imap.starttls()
--- a/rss2email/oauth2.py
+++ b/rss2email/oauth2.py
@@ -1,0 +1,124 @@
+# Copyright (C) 2024 Arun Persaud <apersaud@lbl.gov>
+#
+# This file is part of rss2email.
+#
+# rss2email is free software: you can redistribute it and/or modify it under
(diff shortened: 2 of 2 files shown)


rss2email/email.py
```diff
@@ -254,18 +254,33 @@ def lmtp_send(recipient, message, config=None, section='DEFAULT'):
 def imap_send(message, config=None, section='DEFAULT'):
     if config is None:
         config = _config.CONFIG
-    server = config.get(section, 'imap-server')
-    port = config.getint(section, 'imap-port')
-    _LOG.debug('sending message to {}:{}'.format(server, port))
-    ssl = config.getboolean(section, 'imap-ssl')
-    if ssl:
+    server = config.get(section, "imap-server")
```
auouymous, line 257: Why did you change everything to use double quotes?

rss2email/oauth2.py
```diff
@@ -0,0 +1,124 @@
...
+    client_id: str,
+    client_secret: str,
+    request_url: str = "https://accounts.google.com",
+    scope: str = "https://mail.google.com/",
+) -> None:
+    """Create a reusable oauth2 refresh token.
+
+    Makes two requests to the oauth2 provider to generate a refresh token.
+
+    Refresh tokens are then used to generate authorization tockens for
```
auouymous, line 76: `tockens` typo.

auouymous (changes requested):

auouymous:
> Currently the function needs to be called manually

I haven't fully reviewed the PR yet, but is that function a non-interactive command, or would it open a web browser? Needing to run `r2e refresh-token` via cron every 7 days would be fine, interacting most likely would not.

rss2email/email.py
arunpersaud: I normally automatically run black... will update this now to single quotes

rss2email/oauth2.py
arunpersaud: should be fixed now

rss2email/email.py
arunpersaud: should be fixed now

arunpersaud:
You will only need to generate a refresh token once, since the refresh token can be reused. At the moment you need to call this manually from a python console, but something like 'r2e get-oauth2-refresh-token' would be great. If you have a good idea for a name I can give this a try.

The process currently will take you to google where you OK access for the token and then it redirects you to a non-existing page on localhost. However, that request has the refresh token in it. I believe that some people made a web service one could use as a redirect that then nicely lists the token, but I'm not sure if you think it is OK to links to someones webpage. At the moment the redirect to localhost fails, but if you past the URL into the command line, you get the refresh token printed to the terminal. Not ideal, but at least it doesn't involve any 3rd-party web servers.

auouymous:
When gmail began requiring oauth, the refresh token expired every 7 days, is that no longer the case?

amiryal:
I believe the “right way” to use Gmail for r2e is with [app passwords](https://security.google.com/settings/security/apppasswords). If you must use oauth2, then do not reinvent the wheel; use [the special tool](https://github.com/google/gmail-oauth2-tools/tree/master/go/sendgmail) that mimics regular `sendmail`. (Source: [git-send-email.io](https://web.archive.org/web/20240108134006/https://git-send-email.io/#step-2).)

arunpersaud:
> When gmail began requiring oauth, the refresh token expired every 7 days, is that no longer the case?

I'm not an expert on oauth, but on my computer I got this to work without having to refresh this... so far it's been running for about 1 month. I believe the refresh token last forever and then for every login you generate an authorization token, which is what the code does. Perhaps this is a change from oauth to aouth2?

> I believe the “right way” to use Gmail for r2e is with [app passwords](https://security.google.com/settings/security/apppasswords). If you must use oauth2, then do not reinvent the wheel; use [the special tool](https://github.com/google/gmail-oauth2-tools/tree/master/go/sendgmail) that mimics regular `sendmail`. (Source: [git-send-email.io](https://web.archive.org/web/20240108134006/https://git-send-email.io/#step-2).)

The code I committed does pretty much exactly the same as the python implementation that can be found at git-send-mail. It just combines two steps into into one, e.g. generating the authorization token and generating the string that you need to log in for the IMAP instance of rss2email. You will need some code layer to integrate this with rss2email, and I would think this is pretty much the minimal amount you need. Not sure if it is better to use a 3rd party tool here, but I'm open for whatever is the best way to get this to work in rss2email ;) 

As for app passwords, I'm not familiar with them and don't know how they would be integrated in rss2email. When I go to my google account page (managed by the company I work for), I don't have an option to create app passwords. Also google seems to not recommend them anymore. This is the first text I get when I go to the help for app passwords on my google account "App passwords aren’t recommended and are unnecessary in most cases."

arunpersaud:
Just pushed some small updates and was wondering, if there is a change to get this merged (and if so what else would be needed) or if you rather want to go with 3rd party packages to handle oauth2. Either way, I would be interesting in getting oauth2 support into rss2email ;)

pbrown280:
Hi everyone,

I just got a notice from Google that they will be requiring OAuth starting on Sep 30. I assume that means that my current config will break on that date. And for some reason, I can't seem to enable an app password for the Google account I'm using with rss2email - and even if I could, I'm pretty sure that Google will be sunsetting app passwords eventually, anyway - so the "right" way (paved path) would be to move to OAuth.

So I was wondering if there are plans to get this merged and a new release done in time? Does anyone know? Or would I have to manually patch my setup (which I can certainly do, but I'd hate to break automatic updates)?

Thanks!

Rilomilo (approved):
It's necessary to support OAuth now. Moreover, outlook has deprecated basic auth method.

PR #262, opened 2024-09-15
Avoid sending HTML with too long lines in email body

### LINKED ISSUE #259: Line too long
rss2email 3.14

I have one feed,
http://feeds.macrumors.com/MacRumors-All

that will occasionally error on r2e with this error:

  File "/usr/local/bin/r2e", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/usr/local/lib/python3.12/site-packages/rss2email/main.py", line 204, in run
  File "/usr/local/lib/python3.12/site-packages/rss2email/command.py", line 97, in run
    feed.run(send=args.send, clean=args.clean)
  File "/usr/local/lib/python3.12/site-packages/rss2email/feed.py", line 969, in run
    self._send(sender=sender, message=message)
  File "/usr/local/lib/python3.12/site-packages/rss2email/feed.py", line 918, in _send
    _email.send(recipient=self.to, message=message,
  File "/usr/local/lib/python3.12/site-packages/rss2email/email.py", line 435, in send
    smtp_send(
  File "/usr/local/lib/python3.12/site-packages/rss2email/email.py", line 224, in smtp_send
    smtp.send_message(message, config.get(section, 'from'), recipient.split(','))
  File "/usr/local/lib/python3.12/smtplib.py", line 975, in send_message
    return self.sendmail(from_addr, to_addrs, flatmsg, mail_options,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/smtplib.py", line 897, in sendmail
    raise SMTPDataError(code, resp)
smtplib.SMTPDataError: (500, b'Line too long (see RFC5321 4.5.3.1.6)')

Email "server" is a custom script using aiosmtpd, which seems to require adherence to RFC5321. Would you happen to have any tips on how to avoid this issue?  All I can do now is pause this feed, run r2e run for my other feeds, then r2e run --no-send to skip the bad entries.

Thanks

JulienPalard:
I'm doing so very lazily by always using base64 as the Content-Transfer-Encoding (triggered by the use of UTF-8).

In one hand I'm fixing a bug by removing code which I like, in the other hand it looks disruptive and may cause "trouble" to someone accustomed to things I changed :(

Feels like in 2024 UTF-8 can be always be used ([~98.3 % of internet use UTF-8](https://w3techs.com/technologies/cross/character_encoding/ranking) and what rss2html send is from the internet, I bet the 1.7% left use encodings that encode text that can also be encoded in UTF-8).

Also I don't feel like using html5lib to "properly" wrap HTML, it's a lot of code, so may open a whole lot of other issues. So using base64 is clearly a workaround but without using more code.

I know the messages will be bigger once encoded to base64 instead of just sent as raw ASCII or UTF-8, like 1/3 bigger. I just tested, for a "big" article of 61k bytes of HTML it's 83k bytes of base64, it should sadly not be noticeable in the current internet. If someone feels like it's not OK, one could add an `if`  like "if the body is in ASCII and does not contains long lines, then don't encode it in base64").

This has side effects:
- The use-8bit config is no longer usefull, all mails are sent using 7bit (as the body is base64 encoded it's 7bit proof).
- The 'encodings' configuration option is no longer used neither as UTF-8 is always used.

Closes #259


JulienPalard:
Obviously I broke all the tests :)))

PR #265, opened 2024-11-06
test: run doctest in CI & fix rotten doctest tests

monperrus:
the doctests are not run in CI and a few of them are rotten

this PR fixes this.

going from 42 tests to 57 tests.


monperrus:
:+1:

PR #288, opened 2025-10-03
Respect Cache-Control: max-age header

### LINKED ISSUE #286: openrss.org complains about “Requesting feed content unnecessarily severe”
I am working on making rss2email approved client for openrss.org server, but it has a problem which they mark as SEVERE problem:

> Requesting feed content unnecessarily severe
> 
> This app is unnecessarily re-requesting feed content from websites without waiting until there's actually new content published.
> 
> Why it's a problem
> 
> The unnecessary traffic wastes a website's resources that can be used for its other site visitors.
> 
> What it means for users
> 
> Websites are likely to limit this app from accessing feed content, causing feeds to not update properly or stop working entirely.
> 
> How to fix it
> 
> The application needs to be updated to wait until the max-age of Cache-Control header expires before re-requesting feed content or properly implement conditional requests.

Is there any work on supporting Cache-Control header?

https://openrss.org/guides/developers-guide-to-open-rss-feeds

mcepl:
This change implements support for the `Cache-Control: max-age`
HTTP header to avoid fetching feeds that have not expired.

The `Feed` class now stores `last_checked` and `max_age`
attributes. Before fetching a feed, the application checks if the
cached version is still valid based on these attributes. If the
cache is still fresh, the network request is skipped.

After a successful fetch, the `Cache-Control` header is parsed
from the response, and the `last_checked` and `max_age`
attributes are updated. This ensures that subsequent fetches will
respect the cache duration specified by the feed provider.

This commit adds a test case to verify that the
`Cache-Control: max-age` header is respected.

The test starts a webserver that serves a feed with a `max-age`
value, runs the feed to populate the cache, and then checks that
the feed is not fetched again before the `max-age` expires, and
is fetched again after it expires.

Chore: The commit also includes stylistic changes to the code,
such as using double quotes for strings consistently.

Fixes: https://github.com/rss2email/rss2email/issues/286


test/test_main.py
```diff
@@ -102,11 +154,16 @@ def __init__(self, *args, **kwargs):
...
-            _re.MULTILINE)
-        self.BOUNDARY_REGEXP = _re.compile('===============[^=]+==')
+            r"^User-Agent: rss2email/{0} \({1}\)$".format(
+                _re.escape(_rss2email.__version__), _re.escape(_rss2email.__url__)
+            ),
+            _re.MULTILINE,
+        )
+        self.BOUNDARY_REGEXP = _re.compile("===============[^=]+==")
 
     def clean_result(self, text):
```
amiryal, line 168: Thanks for splitting out the refactoring and `black` commits, really appreciated. Seems like in commit c42e470566935d37fb04ce0b180bd57dfac4c4b5 you forgot to delete this method from here in its old location.

rss2email/feed.py
```diff
@@ -937,40 +986,55 @@ def run(self, send=True, clean=False):
...
+
+        self.last_checked = _time.time()
+        if "cache-control" in parsed.headers:
+            match = _re.search(r"max-age=(\d+)", parsed.headers["cache-control"])
+            if match:
+                self.max_age = int(match.group(1))
+            else:
+                self.max_age = None
+        else:
+            self.max_age = None
```
amiryal, line 1000: This isn’t exactly according to spec. It is certainly an improvement over previous behaviour, but I prefer we read [the spec](https://datatracker.ietf.org/doc/html/rfc9111#name-calculating-age) and implement it correctly from the outset. 


TL;DR: We need to take the value of the Age or Date header into account, and have a distinction between the request time and the response time.

For conforming HTTP/1.1 caches:
```
  response_delay = response_time - request_time;
  corrected_age_value = age_value + response_delay;
```

And in the absence of `age_value`, or just to be safe:
```
  apparent_age = max(0, response_time - date_value);
  corrected_initial_age = max(apparent_age, corrected_age_value);
```

amiryal (changes requested):

rss2email/feed.py
amiryal: Thinking about it again, I realised the proposed implementation is more conservative than the spec, as it assumes an age of zero at response time. So, that’s going to make servers happy, but users may be missing out on timelier updates. Personally I can live with that, so let’s not rush to implement the spec if everyone is cool with the currently proposed implementation.

